PR #5, opened 2020-07-14
Add a way to list all files from a Sorbet project

Morriar:
Using `Spoom::Sorbet::list_files` or the cli command `spoom config files`, one can now list all files belonging to a project that matches Sorbet config.

It takes into account the selected dirs, the ignored strings and the allowed extensions.

### DIFF
--- a/lib/spoom/sorbet/config.rb
+++ b/lib/spoom/sorbet/config.rb
@@ -27,12 +27,13 @@ module Spoom
       extend T::Sig
 
       sig { returns(T::Array[String]) }
-      attr_reader :paths, :ignore
+      attr_reader :paths, :ignore, :allowed_extensions
 
       sig { void }
       def initialize
         @paths = T.let([], T::Array[String])
         @ignore = T.let([], T::Array[String])
+        @allowed_extensions = T.let([], T::Array[String])
       end
 
       class << self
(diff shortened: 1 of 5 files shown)


lib/spoom/sorbet/config.rb
```diff
@@ -77,6 +88,9 @@ def parse_string(sorbet_config)
               if ignore
```
seachel, line 88: If more than one of `ignore`, `extension`, and `skip` are true at the same time due to poor ordering or an accidental extra option in the config file then the "closing" of that flag then I think it will be missed and may mess something up later. Is there any initial verification of the structure of the config file when it's first processed?

test/spoom/cli/commands/config_test.rb
```diff
@@ -98,6 +110,84 @@ def test_display_config_with_ignored_files
             Paths ignored:
              * lib/main
              * test
+
+            Allowed extensions:
+             * .rb (default)
+             * .rbi (default)
+          MSG
+        end
+
```
seachel, line 119: Does it make sense to add tests of the following:

- `--allowed-extension` with extensions after newline
- accidental misuse of lines matching `^--allowed-extension$`, `^--$`, and `^--ignore$` (would block proper processing of next line)
- combinations of `^--allowed-extension$`, `^--$`, and `^--ignore$` in the same config file?

Morriar:
Thanks for your review @seachel.

I replaced the 3 flags by only one that stores the parse `state` so Spoom kind of recover from a bad config file. I also added a test showing the behaviour.

I'm not planing on validating more the file as I let that job to Sorbet.

PR #7, opened 2020-07-22
Add `srb_version` service

Morriar:
Add a service to easily get the version from Sorbet:

```ruby
puts Spoom::Sorbet.srb_version("my_project") # "0.5.5808"
```

It returns `nil` if `srb` is not installed:

```ruby
puts Spoom::Sorbet.srb_version("my_project_without_srb") # nil
```

### DIFF
--- a/test/spoom/sorbet/version_test.rb
+++ b/test/spoom/sorbet/version_test.rb
@@ -1,0 +1,36 @@
+# typed: true
+# frozen_string_literal: true
+
+require "pathname"
+
+require_relative "../cli/cli_test_helper"
+
+module Spoom
+  module Sorbet
+    class VersionTest < Minitest::Test
+      include Spoom::Cli::TestHelper
+      extend Spoom::Cli::TestHelper
+
+      def project_path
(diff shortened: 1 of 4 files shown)


test/spoom/sorbet/version_test.rb
```diff
@@ -0,0 +1,36 @@
...
+            capture_err: true,
+          )
+          assert_nil(version)
+        end
+      end
+
+      def test_return_version_string
+        Bundler.with_clean_env do
+          version = Spoom::Sorbet.srb_version(path: "#{Cli::TestHelper::TEST_PROJECTS_PATH}/project")
+          assert_equal("0.5.5808", version)
```
seachel, line 31: How do you enforce that this version is used? Is it in the config?

Morriar: Yes, it is fixed in the `Gemfile.lock` of the test project: https://github.com/Shopify/spoom/blob/master/test/support/project/Gemfile.lock#L15

We'll change this approach if we have to change the version too often.

PR #8, opened 2020-07-29
Change the strictness from false to true whenever no type checking errors

seachel:
This PR adds a CLI command to change the strictness of a Sorbet sigil from `false` to `true` whenever this change does not create type checking errors. The command is currently called `bump`.

A library is also added help get sigil strictness, check for valid sigils, and change the sigil strictness in strings.

---

Reviewer focus:

- This PR can be read by reviewing the commits
- In the Sigils library, would if be useful to have a verification method available to check that a collection of files have the same strictness level?
- Should any information be logged? e.g. files that are bumped, files that are bumped but have type checking errors?
- Should the term `bump` be replaced?
- Feedback on tests (quality, size, coverage) would also be helpful


lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,39 @@
...
+
+      # returns the full sigil comment string for the passed strictness
+      sig { params(strictness: String).returns(String) }
+      def self.sigil(strictness)
+        "# typed: #{strictness}"
+      end
+
+      # returns true if the passed string is a sigil with valid strictness (else false)
+      sig { params(content: String).returns(T::Boolean) }
+      def self.valid_strictness?(content)
```
Morriar, line 20: If I understand correctly, in this module the semantics are:
* `sigil`: the full `# typed: strictness` comment
* `strictness`: the value of the strictness, ie. `true`, `false`...

Since this method takes the full `sigil` as input, shouldn't be called `valid_sigil?`?

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,39 @@
...
+        VALID_STRICTNESS.include?(strictness(content))
+      end
+
+      # returns the strictness of a sigil in the passed file content string (nil if no sigil)
+      sig { params(content: String).returns(T.nilable(String)) }
+      def self.strictness(content)
```
Morriar, line 26:
```suggestion
      def self.strictness_from_string(sigil)
```

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,39 @@
...
+        matchdata = SIGIL_REGEXP.match(content)
+
+        matchdata[1] if matchdata
```
Morriar, line 29: I believe we can use this form with the safe `&` operator to change it in a oneliner:

```suggestion
        SIGIL_REGEXP.match(content)&.[](1)
```

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,39 @@
...
+  module Sorbet
+    module Sigils
+      extend T::Sig
+
+      VALID_STRICTNESS = ["ignore", "false", "true", "strict", "strong"].freeze
+      SIGIL_REGEXP = /^#\s*typed\s*:\s*(\w*)$/.freeze
+
+      # returns the full sigil comment string for the passed strictness
+      sig { params(strictness: String).returns(String) }
+      def self.sigil(strictness)
```
Morriar, line 14:
```suggestion
      def self.sigil_string(strictness)
```

feels a bit more self-explanatory

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,39 @@
+# typed: true
+# frozen_string_literal: true
+
+module Spoom
+  module Sorbet
+    module Sigils
+      extend T::Sig
+
+      VALID_STRICTNESS = ["ignore", "false", "true", "strict", "strong"].freeze
```
Morriar, line 9: `__STDLIB_INTERNAL` is also a valid strictness even if only used for very specific purposes

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,39 @@
+# typed: true
```
Morriar, line 1: What if we do `spoom bump lib/spoom/sorbet/sigils.rb`, does it bump itself to `strict`? :p

```suggestion
# typed: strict
```

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+  module Cli
+    module Commands
+      class Bump < Base
+        extend T::Sig
+
+        default_task :bump
+
+        desc "bump", "bump Sorbet sigils from `false` to `true` when no errors"
+        sig { params(directory: String, extension: String).returns(T::Array[String]) }
+        def bump(directory = ".", extension = "rb")
```
Morriar, line 19: The default values  for both the `directory` and `extension` should come from the content of `sorbet/config`.
You can use [this method](https://github.com/Shopify/spoom/blob/master/lib/spoom/sorbet.rb#L41) to get all of this done for you.

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+          files_to_bump = Bump.files_with_sigil_strictness(directory, "false", extension)
```
Morriar, line 20: Why not allowing the user to specify the strictness?


```
spoom bump --from ignore --to false
```

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+
+          Bump.change_sigil_in_files(files_to_bump, "true")
```
Morriar, line 22: We should add constants in `Spoom::Sorbet::Sigils` so we don't have to move strings around like this.

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+module Spoom
+  module Cli
+    module Commands
+      class Bump < Base
+        extend T::Sig
+
+        default_task :bump
+
+        desc "bump", "bump Sorbet sigils from `false` to `true` when no errors"
+        sig { params(directory: String, extension: String).returns(T::Array[String]) }
```
Morriar, line 18:
```suggestion
        sig { params(directory: String, extension: String).void }
```

I don't think `thor` commands are supposed to return something

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+        def bump(directory = ".", extension = "rb")
+          files_to_bump = Bump.files_with_sigil_strictness(directory, "false", extension)
+
+          Bump.change_sigil_in_files(files_to_bump, "true")
+
+          output, no_errors = Spoom::Sorbet.srb_tc(File.expand_path(directory), capture_err: true)
```
Morriar, line 24: Can we have a `--force` option that skip the check and bump anyway?

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+
+          return [] if no_errors
+
+          errors = Spoom::Sorbet::Errors::Parser.parse_string(output)
+
+          files_with_errors = Bump.file_names_from_error(errors)
+
+          Bump.change_sigil_in_files(files_with_errors, "false")
```
Morriar, line 32: re logging: Yes, some logging would be useful, we could show the list of files bumped here

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+
+          Bump.change_sigil_in_files(files_to_bump, "true")
+
+          output, no_errors = Spoom::Sorbet.srb_tc(File.expand_path(directory), capture_err: true)
+
+          return [] if no_errors
+
+          errors = Spoom::Sorbet::Errors::Parser.parse_string(output)
+
+          files_with_errors = Bump.file_names_from_error(errors)
```
Morriar, line 30:
```suggestion
          files_with_errors = errors.map(&:file)
```

No need for a method

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+
+          # returns an array of the file names present in the passed
+          sig { params(errors: T::Array[Spoom::Sorbet::Errors::Error]).returns(T::Array[String]) }
+          def self.file_names_from_error(errors)
+            errors.map { |err| err.file }
+          end
+
+          # returns a string containing the strictness of a sigil in a file at the passed path
+          # * returns nil if no sigil
+          sig { params(path: T.any(String, Pathname)).returns(T.nilable(String)) }
```
Morriar, line 61: I know I said to move it here, but let's move it back to the lib. Sorry.

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+          def self.file_strictness(path)
+            content = File.read(path)
+            Spoom::Sorbet::Sigils.strictness(content)
+          end
+
+          # changes the sigil in the file at the passed path to the specified new strictness
+          sig { params(path: T.any(String, Pathname), new_strictness: String).void }
+          def self.change_sigil_in_file(path, new_strictness)
```
Morriar, line 69: Yeah, let's move all of this back to the lib too.

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+            content = File.read(path)
+            File.write(path, Spoom::Sorbet::Sigils.update_sigil(content, new_strictness))
+          end
+
+          # changes the sigil to have a new strictness in a list of files
+          sig { params(path_list: T::Array[T.any(String, Pathname)], new_strictness: String).returns(T::Array[String]) }
+          def self.change_sigil_in_files(path_list, new_strictness)
```
Morriar, line 76: This too

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+          # returns a string containing the strictness of a sigil in a file at the passed path
+          # * returns nil if no sigil
+          sig { params(path: T.any(String, Pathname)).returns(T.nilable(String)) }
+          def self.file_strictness(path)
+            content = File.read(path)
+            Spoom::Sorbet::Sigils.strictness(content)
+          end
+
+          # changes the sigil in the file at the passed path to the specified new strictness
+          sig { params(path: T.any(String, Pathname), new_strictness: String).void }
```
Morriar, line 68: Can we return a boolean showing the if a change has been performed?

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,85 @@
...
+          def self.change_sigil_in_file(path, new_strictness)
+            content = File.read(path)
+            File.write(path, Spoom::Sorbet::Sigils.update_sigil(content, new_strictness))
+          end
+
+          # changes the sigil to have a new strictness in a list of files
+          sig { params(path_list: T::Array[T.any(String, Pathname)], new_strictness: String).returns(T::Array[String]) }
```
Morriar, line 75: Can we only return the list of files changed?

test/spoom/cli/commands/bump_test.rb
```diff
@@ -0,0 +1,159 @@
...
+require_relative "../cli_test_helper"
+
+module Spoom
+  module Cli
+    module Commands
+      class BumpTest < Minitest::Test
+        include Spoom::Cli::TestHelper
+        extend Spoom::Cli::TestHelper
+
+        TEMPORARY_DIRECTORY = "temp"
```
Morriar, line 15:
```suggestion
        TEMPORARY_DIRECTORY = "test-bump"
```

so we're sure it will never conflict with another test

test/spoom/sorbet/sigils_test.rb
```diff
@@ -0,0 +1,229 @@
...
+        assert_equal("# typed: false", sigil)
+      end
+
+      def test_sigil_empty
+        sigil = Sigils.sigil("")
+        assert_equal("# typed: ", sigil)
+      end
+
+      def test_valid_strictness_ignore
+        content  = <<~STR
```
Morriar, line 20: This feels a bit redundant.

Could we do something like:

```ruby
["ignore", "false", "true", "strict", "strong", "   strong   "].each do |strictness|
  content  = <<~STR
    # typed: #{strictness}
    class A; end
  STR
  assert(Sigils.valid_strictness?(content))
end
```

And another one for when it should return false:

```ruby
["", "FALSE", "foo"].each do |strictness|
  content  = <<~STR
    # typed: #{strictness}
    class A; end
  STR
  refute(Sigils.valid_strictness?(content))
end
```

test/spoom/sorbet/sigils_test.rb
```diff
@@ -0,0 +1,229 @@
...
+
+      def test_valid_strictness_none_return_false
+        content  = <<~STR
+          class A; end
+        STR
+
+        refute(Sigils.valid_strictness?(content))
+      end
+
+      def test_strictness_return_ignore
```
Morriar, line 81: Same comment than for `valid_strictness?`, can we loop?

test/spoom/sorbet/sigils_test.rb
```diff
@@ -0,0 +1,229 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module Spoom
+  module Sorbet
+    class SigilsTest < Minitest::Test
+      def test_sigil_nonempty
```
Morriar, line 9: You can also explain the expect behavior in the test name:

```ruby
      def test_sigil_returns_the_sigil_from_a_string
      def test_strictness_returns_nil_if_none_found
      def test_valid_strictness_returns_false_if_the_string_is_empty
```

lib/spoom/cli/commands/bump.rb
Morriar: Or change the `sorbet/config` or your test project to allow `.tmp` as extension. See how it's done here: https://github.com/Shopify/spoom/blob/master/test/spoom/cli/commands/config_test.rb#L124

lib/spoom/cli/commands/bump.rb
Morriar: Yes, same

lib/spoom/cli/commands/bump.rb
Morriar: You can check if `string_before != string_after`

lib/spoom/cli/commands/bump.rb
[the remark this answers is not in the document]
Morriar: Yes, not necessary anymore

lib/spoom/sorbet/sigils.rb
Morriar: I don't think we need the checking services anymore since we only select the files with the relevant sigil.

The checking can then be delegated to `rubocop`.

lib/spoom/sorbet/sigils.rb
Morriar: Right

lib/spoom/sorbet/sigils.rb
seachel: I've changed the method to accept a strictness string and it checks that it's valid. It's used in the initial check on the `--from` and `--to` arguments to the bump command. I'll hold off on any other checks that we don't need now.

lib/spoom/cli/commands/bump.rb
seachel: `change_sigil_in_files` returns the list that's passed originally since all files have their sigil updated. Should I be doing something else in the event of an error when writing the change in the file?

lib/spoom/cli/commands/bump.rb
seachel: Done, thanks.

lib/spoom/sorbet.rb
```diff
@@ -1,6 +1,7 @@
 # typed: true
 # frozen_string_literal: true
 
+require "spoom/sorbet/sigils"
```
Morriar, line 4: Let's follow the alphabetical order

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,61 @@
...
+    module Commands
+      class Bump < Base
+        extend T::Sig
+
+        default_task :bump
+
+        desc "bump", "change Sorbet sigils from one strictness to another when no errors"
+        option :from, type: :string
+        option :to, type: :string
+        option :force, type: :boolean, default: false, aliases: :f
```
Morriar, line 20: We can use `:desc` to provide a description of what this option does.

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,61 @@
...
+        sig { params(directory: String).void }
+        def bump(directory = ".")
+          from = options[:from] ? options[:from] : Sorbet::Sigils::STRICTNESS_FALSE
```
Morriar, line 23: Can we use `Sorbet::Sigils::STRICTNESS_FALSE` as the option `:default` value?

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,61 @@
...
+          to = options[:to] ? options[:to] : Sorbet::Sigils::STRICTNESS_TRUE
```
Morriar, line 24: ditto

lib/spoom/cli/commands/bump.rb
```diff
@@ -0,0 +1,61 @@
...
+
+module Spoom
+  module Cli
+    module Commands
+      class Bump < Base
+        extend T::Sig
+
+        default_task :bump
+
+        desc "bump", "change Sorbet sigils from one strictness to another when no errors"
```
Morriar, line 17:
```suggestion
        desc "bump DIRECTORY", "change Sorbet sigils from one strictness to another when no errors"
```

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      VALID_STRICTNESS = T.let([
+        STRICTNESS_IGNORE,
+        STRICTNESS_FALSE,
+        STRICTNESS_TRUE,
+        STRICTNESS_STRICT,
+        STRICTNESS_STRONG,
+        STRICTNESS_INTERNAL,
+      ].freeze, T::Array[String])
+      SIGIL_REGEXP = T.let(/^#\s*typed\s*:\s*(\w*)\s*$/.freeze, Regexp)
```
Morriar, line 24:
```suggestion

      SIGIL_REGEXP = T.let(/^#\s*typed\s*:\s*(\w*)\s*$/.freeze, Regexp)
```

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,95 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Spoom
+  module Sorbet
+    module Sigils
```
Morriar, line 6: Let's add a short intro to define `sigil` vs `strictness`.

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,95 @@
...
+      sig do
+        params(
+          directory: T.any(String, Pathname),
+          strictness: String,
+          extension: String
+        )
+          .returns(T::Array[String])
+      end
+      def self.files_with_sigil_strictness(directory, strictness, extension = ".rb")
+        # TODO: make sure the result contains unique entries
```
Morriar, line 87: You can `.uniq` your array for that. You can also `.sort` to make sure the output is stable between between platforms.

lib/spoom/cli.rb
```diff
@@ -6,6 +6,7 @@
 require_relative "cli/commands/config"
 require_relative "cli/commands/lsp"
 require_relative "cli/commands/run"
+require_relative "cli/commands/bump"
```
Morriar, line 9:
```suggestion
require_relative "cli/commands/bump"
require_relative "cli/commands/config"
require_relative "cli/commands/lsp"
require_relative "cli/commands/run"
```

lib/spoom/cli/commands/bump.rb
seachel: Does that go above each option? Or only for `:force`?

lib/spoom/cli/commands/bump.rb
seachel: I get this error when I try: `uninitialized constant Spoom::Sorbet::Sigils (NameError)`

lib/spoom/cli/commands/bump.rb
Morriar: at least `:force`

lib/spoom/sorbet/sigils.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+
+      # finds all files in the specified directory with the passed strictness
+      sig do
+        params(
+          directory: T.any(String, Pathname),
+          strictness: String,
+          extension: String
+        )
+          .returns(T::Array[String])
```
Morriar, line 88:
```suggestion
        ).returns(T::Array[String])
```

PR #11, opened 2020-09-23
Spoom displays the count of classes and modules found in the project

Morriar:
Running Spoom now displays the count of classes and modules in the project:

```
$ bundle exec spoom tc metrics
No errors! Great job.
Sigils:
  files: 52
  ignore: 1 (1%)
  true: 46 (88%)
  strict: 5 (9%)

Classes & Modules:
  classes: 276 (including singleton classes)
  modules: 46

Methods:
  methods: 1660
  signatures: 85 (5%)

Sends:
  sends: 2763
  typed: 2183 (79%)
```

Note that the classes count also include singleton classes.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,10 +45,10 @@ GEM
     rubocop-sorbet (0.4.0)
       rubocop
     ruby-progressbar (1.10.1)
-    sorbet (0.5.5803)
-      sorbet-static (= 0.5.5803)
-    sorbet-runtime (0.5.5803)
-    sorbet-static (0.5.5803-universal-darwin-14)
+    sorbet (0.5.5916)
+      sorbet-static (= 0.5.5916)
+    sorbet-runtime (0.5.5916)
+    sorbet-static (0.5.5916-universal-darwin-14)
     thor (1.0.1)
     unicode-display_width (1.7.0)
 
--- a/lib/spoom/sorbet/metrics.rb
+++ b/lib/spoom/sorbet/metrics.rb
@@ -80,7 +80,7 @@ module Spoom
         out.puts "\nClasses & Modules:"
         cc = self['types.input.classes.total']
         cm = self['types.input.modules.total']
-        out.puts "  classes: #{cc}"
+        out.puts "  classes: #{cc} (including singleton classes)"
         out.puts "  modules: #{cm}"
 
         out.puts "\nMethods:"
--- a/lib/spoom/sorbet/metrics.rb
+++ b/lib/spoom/sorbet/metrics.rb
@@ -58,7 +58,7 @@ module Spoom
 
       sig { params(key: String).returns(T.nilable(Integer)) }
       def [](key)
-        metrics[key]
+        metrics[key] || 0
       end
 
       sig { returns(String) }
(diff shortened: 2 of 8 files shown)


Gemfile.lock
```diff
@@ -45,10 +45,10 @@ GEM
     rubocop-sorbet (0.4.0)
       rubocop
     ruby-progressbar (1.10.1)
-    sorbet (0.5.5803)
-      sorbet-static (= 0.5.5803)
-    sorbet-runtime (0.5.5803)
-    sorbet-static (0.5.5803-universal-darwin-14)
+    sorbet (0.5.5916)
```
Morriar, line 48: Version needed for the new classes count metrics

lib/spoom/sorbet/metrics.rb
```diff
@@ -58,7 +58,7 @@ def files_count
 
       sig { params(key: String).returns(T.nilable(Integer)) }
       def [](key)
-        metrics[key]
+        metrics[key] || 0
```
Morriar, line 61: Let's display clean `0` instead of an empty strings

test/spoom/sorbet/metrics_test.rb
```diff
@@ -170,6 +182,10 @@ def test_show_metrics_with_0s
             false: 0
             true: 0
 
+          Classes & Modules:
```
Morriar, line 185: Show that displays `0` correctly if the metrics is not found.

test/spoom/sorbet/version_test.rb
```diff
@@ -28,7 +28,7 @@ def test_return_nil_if_srb_not_installed
       def test_return_version_string
         Bundler.with_clean_env do
           version = Spoom::Sorbet.srb_version(path: "#{Cli::TestHelper::TEST_PROJECTS_PATH}/project")
-          assert_equal("0.5.5808", version)
+          assert_match(/\d\.\d\.\d{4}/, version)
```
Morriar, line 31: As to avoid changing the version every time in the test.

This test will fail if the version does not match `X.X.XXXX`, so it will fail if we cannot read the string and this is what we actually want to check.

test/support/project/Gemfile.lock
```diff
@@ -11,10 +11,10 @@ GEM
   remote: https://rubygems.org/
   specs:
     colorize (0.8.1)
-    sorbet (0.5.5808)
-      sorbet-static (= 0.5.5808)
+    sorbet (0.5.5916)
```
Morriar, line 14: Updated the Sorbet version used in test.

KaanOzkan (approved):
Maybe we should document somewhere (Maybe in output) that singleton classes are also counted.

Morriar:
> Maybe we should document somewhere (Maybe in output) that singleton classes are also counted.

Good idea I added it in the output: ` classes: X (including singleton classes)`.

In the meantime I'm also changing Sorbet so it makes the `singleton_classes` metrics available outside of a debug build so we will be able to display the precise number.

lib/spoom/sorbet/metrics.rb
paracycle: There is a neater trick for this: Initialize the hash for `metrics` with `Hash.new(0)` instead of as `{}`.

PR #12, opened 2020-09-23
Add a --version option

Morriar:
Always useful.

### DIFF
--- a/lib/spoom/cli.rb
+++ b/lib/spoom/cli.rb
@@ -14,6 +14,7 @@ module Spoom
       extend T::Sig
 
       class_option :no_color, desc: "Don't use colors", type: :boolean
+      map T.unsafe(%w[--version -v] => :__print_version)
 
       desc "config", "manage Sorbet config"
       subcommand "config", Spoom::Cli::Commands::Config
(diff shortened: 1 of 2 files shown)


lib/spoom/cli.rb
```diff
@@ -27,6 +28,11 @@ class Main < Thor
       desc "bump", "bump Sorbet sigils from `false` to `true` when no errors"
       subcommand "bump", Spoom::Cli::Commands::Bump
 
+      desc "--version", "show version"
+      def __print_version
```
Morriar, line 32: The `__` prefix make it unaccessible as a command like `spoom __print_version` and avoid having creating the `no_command do .. end` block that Sorbet doesn't like.

lib/spoom/cli.rb
```diff
@@ -14,6 +14,7 @@ class Main < Thor
       extend T::Sig
 
       class_option :no_color, desc: "Don't use colors", type: :boolean
+      map T.unsafe(%w[--version -v] => :__print_version)
```
Morriar, line 17: We map both `--version` and `-v` options to call the `__print_version` method.

We do that a bit differently because we want `-v` to actually behave like a Thor command: `spoom --version` should not be usable with another command. Like `spoom --version tc` makes no sense.

That way we stop the command dispatch.

Morriar:
I also used `-v` since we have no verbose mode yet. This might change in the future and we can reallocate `-v` for it.

paracycle (approved):
Awesome. I've been meaning to add this to `tapioca` as well.

Morriar:
> Awesome. I've been meaning to add this to `tapioca` as well.

Here you go https://github.com/Shopify/tapioca/pull/135.

PR #14, opened 2020-09-24
Improve the README

Morriar:
Document most of the features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ puts Spoom::Sorbet.srb_metrics(capture_err: false)
 
 #### Interacting with LSP
 
-Create a LSP client:
+Create an LSP client:
 
 ```rb
 client = Spoom::LSP::Client.new(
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,6 +163,47 @@ puts config.paths   # "a", "b"
...
+
+Display metrics collected during typechecking:
+
+```ruby
+puts Spoom::Sorbet.srb_metrics(capture_err: false)
+```
+
+#### Interacting with LSP
+
+Create a LSP client:
```
paracycle, line 183:
```suggestion
Create an LSP client:
```

paracycle (approved):
Wow, I hadn't realized that Spoom was doing this much! Amazing work

README.md
Morriar: Dang, I hesitated

paracycle: It really depends on how you expect it to be read. If you expect people to say _El-Es-Pee_ then `an` is appropriate. If it will be read as _Language Server Protocol_ then `a` is fine.

PR #16, opened 2020-09-28
Extract Cli::Commands::Base as a helper module

Morriar:
Some refactoring on how we share code between CLI commands.

This PR extract a new module called `CommandHelper` from what was `Cli::Commands::Base` so the code can be shared without inheriting anything and we don't need the `no_commands` block anymore.


KaanOzkan (approved):
This is just favoring inclusion over inheritance right?

PR #18, opened 2020-09-28
Clean some unused metrics fields

Morriar:
Those metrics are only initialized if you pass specific options to Sorbet:

```
      --metrics-branch branch   Branch to report in metrics export (default:
                                none)
      --metrics-sha sha1        Sha1 to report in metrics export (default:
                                none)
      --metrics-repo repo       Repo to report in metrics export (default:
                                none)
      --metrics-extra-tags key1=value1,key2=value2
                                Extra tags to report, comma separated
                                (default: "")
```

We actually don't care about them so let's remove the fields.

### DIFF
--- a/test/spoom/sorbet/metrics_test.rb
+++ b/test/spoom/sorbet/metrics_test.rb
@@ -33,10 +33,6 @@ module Spoom
            ]
           }
         ERR
-        assert_equal("MyRepo", metrics.repo)
-        assert_equal("1234", metrics.sha)
-        assert_equal("master", metrics.branch)
-        assert_equal("Success", metrics.status)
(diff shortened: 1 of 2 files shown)


test/spoom/sorbet/metrics_test.rb
```diff
@@ -33,10 +33,6 @@ def test_parses_metrics
            ]
           }
         ERR
-        assert_equal("MyRepo", metrics.repo)
```
Morriar, line 36: I let the data lines 18 to 34 as to show that it doesn't break if the output actually contains those fields.

paracycle (approved):

PR #20, opened 2020-09-29
Define a clean printers abstraction

Morriar:
This pull-request refactor `Spoom::FileTree` and `Spoom::Cli::SymbolTreePrinter` so they use the same abstraction and share options.

No functional change.


lib/spoom/cli/commands/lsp.rb
```diff
@@ -127,6 +122,21 @@ def lsp_client
             client
           end
 
+          def print_list(list)
+            printer = Spoom::LSP::SymbolPrinter.new(indent_level: 1, colors: !options["no_color"])
```
Morriar, line 126: `indent_level: 1` as to preserve the previous output format.

lib/spoom/sorbet/lsp/base.rb
```diff
@@ -56,3 +56,13 @@ def initialize(method, params)
     end
   end
 end
+
+class String
```
Morriar, line 60: Moved here from the deleted `symbol_printer.rb` file.

lib/spoom/cli/symbol_printer.rb
```diff
@@ -1,71 +0,0 @@
...
-      def self.print_object(object, no_color)
-        printer = SymbolPrinter.new(2, no_color)
-        printer.visit(object)
-      end
-
-      def initialize(default_indent, no_color = false)
-        @seen = Set.new
-        @current_indent = default_indent
-        @no_color = no_color
-        String.disable_colorization = no_color
```
Morriar, line 29: Removing this is the big win: it disabled coloring at the app level.

This was a hack when I moved the LSP prototype to `spoom`.

We don't need it anymore.

Morriar:
Note: I plan to apply the same thing to `Spoom::Sorbet::Metrics.show` at some point but it need some more refactoring first.

lib/spoom/cli/commands/lsp.rb
```diff
@@ -127,6 +122,21 @@ def lsp_client
...
+            list.each do |item|
+              printer.printt
+              printer.print("* ")
+              printer.print_object(item)
+              printer.printn
+            end
+          end
+
+          def print_object(object)
+            printer = Spoom::LSP::SymbolPrinter.new(indent_level: 2, colors: !options["no_color"])
```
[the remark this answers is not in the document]
Morriar: Good call, `print_object` was called multiple time from the same command. I added a commit cleaning all of this (I finally had to change the test output).

PR #21, opened 2020-09-29
Rename option --no-color => --color

Morriar:
As to avoid it showing as `--no-no-color` in Thor help.

I also added a test showing the output.


paracycle (approved):
I like it, makes the code much cleaner.

PR #22, opened 2020-09-29
Introduce Spoom::Snapshot

Morriar:
In my quest to migrate parts of `sorbet-metrics` into `spoom` I will have to collect and store more metrics than what `sorbet --metrics-file` file produces.

This PR prepares the land for this:
* Demote `Sorbet::Metrics` as a mere hash (without the other fields, the data are homogeneous, no need of a struct for that)
* Introduce an abstraction on top of the hash called `Snapshot` that can hold more than integers
* Use the Printer abstract to display the snapshot
* Promote `spoom tc metrics` as a to level command `spoom metrics`


lib/spoom/snapshot.rb
```diff
@@ -0,0 +1,96 @@
...
+    V = T.type_alias { type_member(fixed: T.untyped) }
+    Elem = type_member(fixed: String)
+
+    sig { params(path: String).returns(Snapshot) }
+    def self.snapshot(path: '.')
+      snapshot = Snapshot.new
+      metrics = Spoom::Sorbet.srb_metrics(path: path, capture_err: false)
+      return snapshot unless metrics
+
+      snapshot["user_time"] = metrics["run.utilization.user_time.us"]
```
Morriar, line 19: I hand pick the data so the resulting JSON (later) will be smaller (we don't care about a lot of the data generated by `sorbet --metrics`

lib/spoom/cli/commands/metrics.rb
```diff
@@ -0,0 +1,26 @@
...
+      class Metrics < Thor
+        include Spoom::Cli::CommandHelper
+
+        default_task :snapshot
+
+        desc "snapshot", "run srb tc and display metrics"
+        def snapshot
+          in_sorbet_project!
+
+          snapshot = Spoom::Snapshot.snapshot
```
paracycle, line 20: Why is this class called `Snapshot`? It seems to be representing `Metrics` to me.

lib/spoom/snapshot.rb
```diff
@@ -0,0 +1,96 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Spoom
+  class Snapshot < Hash
```
paracycle, line 5: I believe it is fine here, but in general it is bad practice to subclass from a built-in type.

lib/spoom/snapshot.rb
```diff
@@ -0,0 +1,96 @@
...
+    extend T::Sig
+    extend T::Generic
+
+    K = type_member(fixed: String)
+    V = T.type_alias { type_member(fixed: T.untyped) }
```
paracycle: Whoa, what is going on here?

lib/spoom/sorbet.rb
```diff
@@ -55,7 +55,7 @@ def self.srb_version(*arg, path: '.', capture_err: false)
       out.split(" ")[2]
     end
 
-    sig { params(arg: String, path: String, capture_err: T::Boolean).returns(T.nilable(Metrics)) }
+    sig { params(arg: String, path: String, capture_err: T::Boolean).returns(T.nilable(T::Hash[String, Integer])) }
```
Morriar, line 58: `Metrics` was a struct with only one field: the hash.

Let's remove the middleman.

lib/spoom/sorbet/metrics.rb
```diff
@@ -1,90 +1,32 @@
...
-      sig { params(out: T.any(IO, StringIO)).void }
-      def show(out = $stdout)
-        files = files_count
-
-        out.puts "Sigils:"
-        out.puts "  files: #{files}"
-        files_by_strictness.each do |sigil, value|
-          next unless value && value > 0
-          out.puts "  #{sigil}: #{value}#{percent(value, files)}"
+        obj["metrics"].each_with_object(Hash.new(0)) do |metric, metrics|
```
Morriar, line 25: Basically the only thing remaining here is stripping the keys prefix so it's easier to use in the rest of `spoom`.

lib/spoom/snapshot.rb
```diff
@@ -0,0 +1,96 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Spoom
+  class Snapshot < Hash
+    extend T::Sig
+    extend T::Generic
+
+    K = type_member(fixed: String)
+    V = type_member(fixed: T.untyped)
```
Morriar, line 10: Untyped so we can store the version, sha, or any other data in the snapshot

lib/spoom/snapshot.rb
Morriar, line 5: My first try was using a typed struct.

But since the snapshot data are mainly used for display it means no looping over the fields etc without meta-programming.

I opted to use a hash instead to make it easier for the client.

paracycle: There is always `T::Struct#serialize` method to turn it into a Hash. That might still be better, don't you think?

lib/spoom/snapshot.rb
```diff
@@ -0,0 +1,96 @@
...
+    Elem = type_member(fixed: T.untyped)
```
paracycle, line 11: Why is this needed?

Morriar:
Reroll: replace the `Hash` subclass by a `T::Struct`.

lib/spoom/cli/commands/metrics.rb
Morriar: Metrics generally means something you can count but I also want to store things like version numbers for example.

The next pull-request introduce the concept of `Timeline`, each set of data collected of a commit is a `Snapshot`.

Also it was confusing between `Spoom::Metrics` (the snapshot) and `Spoom::Sorbet::Metrics` (the metrics produced by Sorbet).

paracycle: WDYT about `MetricSnapshot` or `MetricsSnapshot`. It more clearly defines what this is a snapshot of.

Morriar: This will become clearer as I add data to this thing in the following pull-requests.

paracycle (approved):

Morriar:
@paracycle finally I nested `Snapshot` inside `Spoom::Coverage`.

This will give us:

```sh
spoom coverage [snapshot]
spoom coverage timeline # to collect multiple snapshots
spoom coverage report # to generate the report
spoom coverage open # to open the report

spoom coverage files # might as well move it here too?
```

KaanOzkan (approved):

lib/spoom/snapshot.rb
```diff
@@ -0,0 +1,105 @@
...
+  module Coverage
+    extend T::Sig
+
+    sig { params(path: String).returns(Snapshot) }
+    def self.snapshot(path: '.')
+      snapshot = Snapshot.new
+      metrics = Spoom::Sorbet.srb_metrics(path: path, capture_err: true)
+      return snapshot unless metrics
+
+      snapshot.files = metrics["types.input.files"] || 0
```
paracycle, line 14: Super nitpicky but I think `fetch` is preferred in these cases:
```suggestion
      snapshot.files = metrics.fetch("types.input.files", 0)
```

lib/spoom/sorbet/metrics.rb
```diff
@@ -1,90 +1,32 @@
 # typed: strict
 # frozen_string_literal: true
 
+require_relative "sigils"
+
 module Spoom
   module Sorbet
-    class Metrics < T::Struct
+    module Metrics
```
paracycle, line 8: If the only responsibility is to parse the metrics output, how about:
```suggestion
    module MetricsParser
```

paracycle (approved):
This version makes much more sense. Two trivial suggestions, nothing fundamental.

PR #23, opened 2020-09-30
Add `spoom coverage timeline`

Morriar:
This command basically run `spoom coverage snapshot` for each month your project has existed.

It finds the first commit sorbet was introduced, then selects one commit by month until now to build the timeline.
This can be controlled with the `--from` and `--to` options.

The `--save-dir` option allows the user the save the timeline in JSON format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,24 @@ Show metrics about the project contents and the typing coverage:
 $ spoom coverage
 ```
 
+Show typing coverage evolution based on the commits history:
+
+```
+$ spoom coverage timeline
+```
+
+Show typing coverage evolution based on the commits history between specific dates:
+
+```
+$ spoom coverage timeline --from YYYY-MM-DD --to YYYY-MM-DD
+```
+
+Save the typing coverage evolution as JSON in a specific directory:
+
+```
+$ spoom coverage timeline --save-dir data/
+```
+
 #### Change the sigil used in files
 
 Bump the strictness from all files currently at `typed: false` to `typed: true` where it does not create typechecking errors:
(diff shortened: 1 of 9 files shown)


paracycle (approved):
This looks good, but shouldn't we be worried about changing Sorbet versions between commits?

Morriar:
> This looks good, but shouldn't we be worried about changing Sorbet versions between commits?

We have two options:

1. Using the right Sorbet version for the commit so we don't have errors. But that also means the metrics file format and content will not be the same, sorbet options not the same etc. Plus we need to `bundle install` each version of the app we snapshot.

2. Using the same version for all commit: this might create errors for breaking changes like we've seen with `T.type_alias`. But we ignore the errors so it works, most of the numbers will be accurate, some will not.

I think we can make this an option later?

paracycle:
> 1. Using the right Sorbet version for the commit so we don't have errors. But that also means the metrics file format and content will not be the same, sorbet options not the same etc. Plus we need to `bundle install` each version of the app we snapshot.

We don't necessarily need `bundle install`, we just need to be able to `gem install sorbet-static -v <sorbet_version>` and reach into its `libexec` folder to run `sorbet` as an executable. Agreed on the metric format difference but maybe we can compensate for it in `spoom` back to some specific version.

> 2. Using the same version for all commit: this might create errors for breaking changes like we've seen with `T.type_alias`. But we ignore the errors so it works, most of the numbers will be accurate, some will not.

Sure, that works for me.

> I think we can make this an option later?

I don't mind it being an option for later, just wanted to raise the point.

PR #24, opened 2020-09-30
Speed up the bump task

ghiculescu:
`srb tc` could return many errors from the same file. Only change sigil once per file.


Morriar (approved):
Good catch thanks!

KaanOzkan (approved):

paracycle (approved):

PR #25, opened 2020-10-01
Bump: ignore paths that don't exist

ghiculescu:
This task was crashing in our code base, due to an extremely large `srb tc` error that wasn't parsed correctly.

I can't copy the error trace because it's basically an entire file, but the gist is that it's trying to wrap that whole class in `T.let(KLASS_NAME...end, T.untyped)`. Anyway, the filename is parsing totally wrong. To avoid errors I'm just ignoring paths that aren't really file paths.

### DIFF
--- a/lib/spoom/cli/commands/bump.rb
+++ b/lib/spoom/cli/commands/bump.rb
@@ -48,8 +48,8 @@ module Spoom
           errors = Sorbet::Errors::Parser.parse_string(output)
 
           files_with_errors = errors.map do |err|
-            File.join(directory, err.file)
-          end.compact.select {|path| File.file?(path)}
+            File.join(directory, err.file) if File.file?(err.file)
+          end.compact
 
           Sorbet::Sigils.change_sigil_in_files(files_with_errors, from)
         end
--- a/lib/spoom/cli/commands/bump.rb
+++ b/lib/spoom/cli/commands/bump.rb
@@ -48,7 +48,8 @@ module Spoom
           errors = Sorbet::Errors::Parser.parse_string(output)
 
           files_with_errors = errors.map do |err|
-            File.join(directory, err.file) if File.file?(err.file)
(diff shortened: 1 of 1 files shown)


Morriar:
👋 @ghiculescu thanks for catching this!

Do you mind adding a test that reproduces the problem?
It doesn't needs to be realistic, just a string that will produce the behaviour.

In the mean time I'll take a look at Sorbet see if we can add an option to remove some of the output so it's easier to use from tools. The LSP could be a good approach too.

ghiculescu:
I can have a look at adding a test later today, but while I'm here. One of the items in the array returned by `Sorbet::Errors::Parser.parse_string(output)`  had file = `end, T.untyped)`.

ghiculescu:
Actually, this should do it.

```
# typed: true
require "test_helper"

class TheTest < ActiveSupport::TestCase
  test "foo" do
    q = something do
      q = something_else.new
    end
  end
end
```

This is the problematic error:

```
test/models/platform/test.rb:7: Changing the type of a variable in a loop is not permitted https://srb.help/7001
     7 |      q = something_else.new
                  ^^^^^^^^^^^^^^^^^^
  Existing variable has type: NilClass
  Attempting to change type to: T.untyped

  Autocorrect: Use `-a` to autocorrect
    test/models/platform/test.rb:4: Replace with T.let(class TheTest < ActiveSupport::TestCase
  test "foo" do
    q = something do
      q = something_else.new
    end
  end
end, T.untyped)
     4 |class TheTest < ActiveSupport::TestCase
     5 |  test "foo" do
     6 |    q = something do
     7 |      q = something_else.new
     8 |    end
     9 |  end
    10 |end
```

Morriar:
@ghiculescu I moved your commit to this pull-request so I could add the test: https://github.com/Shopify/spoom/pull/28

Closing this one, thank you!

PR #26, opened 2020-10-06
Clean test helpers

Morriar:
This pull-request introduces `TestProject`, an abstraction to build temporary projects so we can test `spoom`.

We currently rely on the projects inside `test/support` to run the `spoom` client tests which as many limitations:
* a lot of boiler plate content just for a few Ruby files we actually test
* adding a new test project is hard
* touching an existing project might impact other tests (like metrics or LSP)
* any change to the project from a test must be undone

With the `TestProject` abstraction each test can build a custom project with only what is needed for testing.

This simplifies testing of edge-cases (outside of `git` dir, no `sorbet/config` etc...) and puts the tested code closer to the testing code which make the overall result cleaner and easier to read/understand.

This also removes a lot of helpers related to testing.


KaanOzkan (approved):
I like this abstraction and the api 👍

paracycle (approved):
Neat

PR #27, opened 2020-10-06
Add `--path` option

Morriar:
This pull-request adds a `--path` option to `spoom` so it can run in a different directory.

Why is it useful?
* When working on new features, one can use `--path` to test the feature on different projects without having to change their Gemfiles.
* Spoom can be used as a system wide gem and ran on projects that does not use spoom (some commands even work if the project doesn't have sorbet installed).

I also added a few tests ensuring each command works correctly with the `--path` option.

### DIFF
--- a/lib/spoom/sorbet/lsp.rb
+++ b/lib/spoom/sorbet/lsp.rb
@@ -11,10 +11,12 @@ require_relative 'lsp/errors'
 module Spoom
   module LSP
     class Client
-      def initialize(sorbet_cmd, *sorbet_args)
+      def initialize(sorbet_cmd, *sorbet_args, path: ".")
         @id = 0
         Bundler.with_clean_env do
-          @in, @out, @err, @status = Open3.popen3([sorbet_cmd, *sorbet_args].join(" "))
+          opts = {}
+          opts[:chdir] = path
+          @in, @out, @err, @status = Open3.popen3([sorbet_cmd, *sorbet_args].join(" "), opts)
         end
       end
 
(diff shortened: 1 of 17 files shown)


KaanOzkan (approved):

lib/spoom/sorbet/lsp.rb
```diff
@@ -11,10 +11,12 @@
 module Spoom
   module LSP
     class Client
-      def initialize(sorbet_cmd, *sorbet_args)
+      def initialize(sorbet_cmd, *sorbet_args, path: ".")
         @id = 0
         Bundler.with_clean_env do
-          @in, @out, @err, @status = Open3.popen3([sorbet_cmd, *sorbet_args].join(" "))
+          opts = {}
+          opts[:chdir] = path
```
rafaelfranca, line 18:
```suggestion
          opts = { chdir: path }
```

?

Morriar: This one is to please Sorbet which has a hard time figuring how to pass the hash:

https://sorbet.run/#%23%20typed%3A%20true%0A%0AOpen3.capture3(%22ls%22%2C%20%7Bchdir%3A%20%22.%22%7D)%0A%0Aopts%20%3D%20%7Bchidr%3A%20%22.%22%7D%0AOpen3.capture3(%22ls%22%2C%20opts)%0A%0Aopts%20%3D%20%7B%7D%0Aopts%5B%3Achidr%5D%20%3D%20%22.%22%0AOpen3.capture3(%22ls%22%2C%20opts)

 This is a known bug and they recently started working on it, should be resolved once we implement Ruby2.7 kargs changes correctly.

PR #29, opened 2020-10-07
Improve error line matching

paracycle:
### Motivation 

By making sure that we can match the whole line as a proper error line, we can decrease the chances of matching lines that look like error lines, since they are not indented. This can prevent cases like #25 happening in the first place.

## Implementation

This PR introduces a regular expression that matches an error line completely.

In general, this could be much more expensive to do, however, we already had the code to match if a line started with an empty space or not. So, by prefixing the beginning of our error line match to be a non-whitespace character, we can save on scanning the rest of the string for a match. Thus, the full regexp match will only execute on non-indented lines.

The added benefit is that once we match a line to be an error line, we can match all components of it in a single pass and construct the `Error` object from those parts.

## Tests

I've included a test that includes various weird file names and edge-cases to ensure that the error line parsing can handle those properly.

### DIFF
--- a/lib/spoom/sorbet/errors.rb
+++ b/lib/spoom/sorbet/errors.rb
@@ -16,6 +16,19 @@ module Spoom
           "or set SORBET_SILENCE_DEV_MESSAGE=1 in your shell environment.",
         ]
 
+        ERROR_LINE_MATCH_REGEX = %r{
+          ^         # match beginning of line
+          (\S[^:]*) # capture filename as something that starts with a non-space character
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
Thanks for explaining the regex. `#captures`  makes it very tidy too.

Morriar (approved):
Amazing, thanks!

rafaelfranca (approved):

paracycle:
I added a small change to match single character file names, too. Don't think there would be many of them, but better safe than sorry.

PR #30, opened 2020-10-07
Add spoom coverage timeline --bundle-install option

Morriar:
As discussed in #23 with @paracycle we should provide a way to collect the snapshot data from an installed bundle.

This PR introduces `spoom coverage timeline --bundle-install` which will run `bundle install` for each commit in the timeline before running Sorbet to collect the metrics.

If a `--bundle-install` is used and bundle install doesn't finishes properly, the commit is skipped.

### DIFF
--- a/lib/spoom/cli/commands/coverage.rb
+++ b/lib/spoom/cli/commands/coverage.rb
@@ -25,6 +25,7 @@ module Spoom
         option :from, type: :string
         option :to, type: :string, default: Time.now.strftime("%F")
         option :save_dir, type: :string
+        option :bundle_install, type: :boolean, desc: "Execute `bundle install` before collecting metrics"
         def timeline
           in_sorbet_project!
           path = exec_path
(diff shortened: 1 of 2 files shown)


paracycle (approved):
👌

lib/spoom/cli/commands/coverage.rb
```diff
@@ -25,6 +25,7 @@ def snapshot
         option :from, type: :string
         option :to, type: :string, default: Time.now.strftime("%F")
         option :save_dir, type: :string
+        option :bundle_install, type: :boolean, desc: "Execute `bundle install` before collecting metrics"
```
KaanOzkan, line 28: It may be beneficial here to give an explanation about why a user would want to specify this: Talking about how this "fixes potential errors due to difference in Sorbet versions".

lib/spoom/cli/commands/coverage.rb
```diff
@@ -90,6 +102,18 @@ def parse_date(string, option)
             say_error("Invalid date `#{string}` for option #{option} (expected format YYYY-MM-DD)")
             exit(1)
           end
+
+          def bundle_install(path, sha)
+            opts = {}
+            opts[:chdir] = path
+            out, status = Open3.capture2e("bundle install", opts)
+            unless status.success?
+              say_error("Can't run `bundle install` for commit #{sha}")
```
KaanOzkan, line 111: I think we can give more info here to say something like 

```
"Can't run `bundle install` for commit #{sha}. Coverage snapshot is skipped"
```

lib/spoom/cli/commands/coverage.rb
Morriar: Good point, I added a not in the `README`.

lib/spoom/cli/commands/coverage.rb
Morriar: Done.

KaanOzkan (approved):

PR #31, opened 2020-10-07
Save sorbet-version used in snapshots

Morriar:
Spoom now provides two ways to get the Sorbet version used:

1. Through `Spoom::Sorbet.srb_version` which runs `bundle exec sorbet --version`.
2. Through `Spoom::Sorbet.srb_version_from_gemfile_lock` which parses the content of the `Gemfile.lock` for the current commit.

This makes us able to save the sorbet version used for a snapshot for the two cases we have in `spoom timeline`:

1. We don't use `--bundle-install` so we read the version from the `Gemfile.lock` (useful when running the timeline from another context, like we do in `sorbet-metrics` for core).
2. We use `--bundle-install` so we can just ask Sorbet.

### DIFF
--- a/lib/spoom/cli/commands/coverage.rb
+++ b/lib/spoom/cli/commands/coverage.rb
@@ -17,7 +17,9 @@ module Spoom
         def snapshot
           in_sorbet_project!
 
-          snapshot = Spoom::Coverage.snapshot(path: exec_path)
+          path = exec_path
+          snapshot = Spoom::Coverage.snapshot(path: path)
+          snapshot.sorbet_version = Spoom::Sorbet.srb_version(path: path)
           snapshot.print
         end
 
--- a/lib/spoom/cli/commands/coverage.rb
+++ b/lib/spoom/cli/commands/coverage.rb
@@ -77,9 +77,11 @@ module Spoom
               Bundler.with_clean_env do
                 next unless bundle_install(path, sha)
                 snapshot = Spoom::Coverage.snapshot(path: path)
+                snapshot.sorbet_version = Spoom::Sorbet.srb_version(path: path) if snapshot
               end
             else
               snapshot = Spoom::Coverage.snapshot(path: path)
+              snapshot.sorbet_version = Spoom::Sorbet.srb_version_from_gemfile_lock(path: path)
             end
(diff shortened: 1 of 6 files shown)


lib/spoom/cli/commands/coverage.rb
```diff
@@ -17,7 +17,9 @@ class Coverage < Thor
         def snapshot
           in_sorbet_project!
 
-          snapshot = Spoom::Coverage.snapshot(path: exec_path)
+          path = exec_path
+          snapshot = Spoom::Coverage.snapshot(path: path)
+          snapshot.sorbet_version = Spoom::Sorbet.srb_version(path: path)
```
paracycle, line 22: Super nitpicky, but I think I would prefer it more if the API was just:
```suggestion
          snapshot = Spoom::Coverage.snapshot(path: path)
```
and the snapshot process knew how to get the srb version by default.

lib/spoom/cli/commands/coverage.rb
```diff
@@ -75,9 +77,11 @@ def timeline
               Bundler.with_clean_env do
                 next unless bundle_install(path, sha)
                 snapshot = Spoom::Coverage.snapshot(path: path)
+                snapshot.sorbet_version = Spoom::Sorbet.srb_version(path: path) if snapshot
               end
             else
               snapshot = Spoom::Coverage.snapshot(path: path)
+              snapshot.sorbet_version = Spoom::Sorbet.srb_version_from_gemfile_lock(path: path)
```
paracycle, line 84: In this case, maybe we can optionally flag that the version should be parsed from the gemfile:
```suggestion
              snapshot = Spoom::Coverage.snapshot(path: path, version_source: :gemfile)
```
kind of thing.

lib/spoom/cli/commands/coverage.rb
Morriar: We can always read the version from the `Gemfile.lock` and get this done directly in the snapshot 👍

Morriar:
@paracycle I changed the snapshot so it always reads the version from the Gemfile

paracycle (approved):
👍

PR #32, opened 2020-10-07
Add `--save` to `spoom coverage`

Morriar:
This PR does two things: 

1. First, this PR splits the current `--save-dir PATH` into two options: `--save` to ask for save and `--save-dir` to specify the path. This is because I'm tired of typing `--save-dir PATH` when I just want to save to the default dir.

2. Then it reuses these two options on `spoom coverage snapshot` so I don't have to run `spoom coverage snapshot --from F --to T` just to save the snapshot on the current commit (this also makes it easier to get the upcoming `spoom report` feature working with default dirs).

### DIFF
--- a/lib/spoom/cli/commands/coverage.rb
+++ b/lib/spoom/cli/commands/coverage.rb
@@ -14,12 +14,24 @@ module Spoom
         default_task :snapshot
 
         desc "snapshot", "run srb tc and display metrics"
+        option :save, type: :boolean, desc: "Save snapshot data as json"
+        option :save_dir, type: :string, desc: "Save json data under a specific directory", default: "spoom_data"
         def snapshot
           in_sorbet_project!
 
           path = exec_path
           snapshot = Spoom::Coverage.snapshot(path: path)
           snapshot.print
+
+          save_dir = options[:save] ? options[:save_dir] : nil
+          return unless save_dir
+
+          FileUtils.mkdir_p(save_dir)
+          name = snapshot.commit_sha
+          name = Time.now.getutc.to_i unless name
+          file = "#{save_dir}/#{name}.json"
+          File.write(file, snapshot.serialize.to_json)
+          puts "\nSnapshot data saved under #{file}"
         end
 
         desc "timeline", "replay a project and collect metrics"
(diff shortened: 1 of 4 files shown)


KaanOzkan (approved):
I don't love the usage but at the same time I don't think it's worth changing how we deal with command line options now. Ideally I'd prefer

`spoom coverage --save=my_data/` or something similar that doesn't require both options. But again it's not important 😄

Morriar:
> `spoom coverage --save=my_data/` or something similar that doesn't require both options. But again it's not important 😄

That's  exactly the idea.

This PRs basically sets the default path for spoom data as `spoom_data/` so all options work without the user needing to pass `--save-dir`, only `--save`.

The option `--save-dir` is more for scripts (ie `sorbet-metrics`) who want their data somewhere else.

So for normal users you just do `spoom coverage --save`.

KaanOzkan:
Yeah that's true normal users won't use this which makes what I suggested less important.

Although I suggested that for the use case of specifying an explicit save directory. Looking at here: https://github.com/Shopify/spoom/pull/32/files#diff-04c6e90faac2675aa89e2176d2eec7d8R94 it felt off to use 2 flags to achieve this. And the fact that `save-dir` flag on it's own without `save` wouldn't work (?)

`--save=/path/to/dir` is more intuitive to me. Again, not important.

Morriar:
@KaanOzkan here you go, the two options are now merged 👍

lib/spoom/cli/commands/coverage.rb
```diff
@@ -11,21 +11,21 @@ module Commands
...
         def snapshot
           in_sorbet_project!
 
           path = exec_path
           snapshot = Spoom::Coverage.snapshot(path: path)
           snapshot.print
 
-          save_dir = options[:save] ? options[:save_dir] : nil
+          save_dir = options[:save]
           return unless save_dir
```
KaanOzkan, line 28: I don't think we need this anymore. I don't know if you want to still keep it.

KaanOzkan (approved):

lib/spoom/cli/commands/coverage.rb
Morriar: It's not here anymore: 
image

KaanOzkan: Wait I'm confused now 😄 

I meant I don't think we need `return unless save_dir`. Since `save_dir` will always exist due to the default value right? I assume that's how `lazy_default` works.

Morriar: With `lazy_default`, `options[:save]` returns:
* `X` if `--save X` was passed
* the lazy default value if `--save` was passed
* and `nil` otherwise

PR #35, opened 2020-10-09
Allow Snapshot to be saved/loaded with JSON

Morriar:
Also add to metrics to the snapshot:
* timestamp
* sorbet run time

And a test to ensure serialization, deserialization works fine.

### DIFF
--- a/lib/spoom/snapshot.rb
+++ b/lib/spoom/snapshot.rb
@@ -39,6 +39,7 @@ module Spoom
     class Snapshot < T::Struct
       extend T::Sig
 
+      prop :timestamp, Integer, default: Time.new.getutc.to_i
       prop :sorbet_version, T.nilable(String), default: nil
       prop :duration, Integer, default: 0
       prop :commit_sha, T.nilable(String), default: nil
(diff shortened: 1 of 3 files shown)


lib/spoom/snapshot.rb
```diff
@@ -23,6 +23,9 @@ def self.snapshot(path: '.')
       snapshot.calls_typed = metrics.fetch("types.input.sends.typed", 0)
       snapshot.calls_untyped = metrics.fetch("types.input.sends.total", 0) - snapshot.calls_typed
 
+      snapshot.duration += metrics.fetch("run.utilization.system_time.us", 0)
+      snapshot.duration += metrics.fetch("run.utilization.user_time.us", 0)
```
KaanOzkan, line 27: I just wanted to point out that if the program runs on multiple CPUs `sys` + `user` time won't be equal to program duration. I don't know enough Ruby internals and if parts of the program will run in parallel with the GIL.

KaanOzkan (approved):

lib/spoom/snapshot.rb
Morriar: I agree and it's not to be used as a benchmark tool.

The idea is more to spot big changes in typechecking time:

image

PR #36, opened 2020-10-09
Manually parse version from Gemfile.lock

Morriar:
It turns out that you can't even _parse_ a Gemfile.lock from another version Bundle version (it raises).

This PR gets the version from the `Gemfile.lock` manually instead.

I also added a way to specify which gem you want the version from, can be useful to collect the info about `sorbet-runtime` (see test).

### DIFF
--- a/lib/spoom/snapshot.rb
+++ b/lib/spoom/snapshot.rb
@@ -28,7 +28,7 @@ module Spoom
         snapshot.sigils[strictness] = T.must(metrics["types.input.files.sigil.#{strictness}"])
       end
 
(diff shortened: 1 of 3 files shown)


KaanOzkan (approved):

paracycle:
What raises exactly? The `Gemfile.lock` file format does not really change between versions, so it should be parsable across versions. What was the error you got?

Morriar:
The parser just refuses to parse it voluntarily: https://github.com/rubygems/bundler/blob/master/lib/bundler/lockfile_parser.rb#L94-L109.

PR #38, opened 2020-10-09
Introduce HTML coverage report

Morriar:
This PR reuse all the features from snapshot, timeline and coverage to create a beautiful report.

Here's how to generate the report for `spoom`

```
cd spoom
spoom coverage timeline --save
spoom coverage report
spoom coverage open
```

And here's the output:

image
image
image
image


Morriar:
Also the map is interactive 😄

paracycle (approved):
This is amazing progress. 🎉  However, I think we should evolve this so that the Ruby generation part and the HTML/JS/CSS part have a well defined interface for what and how data is going to be passed between them. If we do that, we should be able to edit each component independently and open the door for users to replace templates with their own things, if they want.

Morriar:
New version using Erb templates for the main HTML components.

I didn't use `Thor::actions` as to not generate files all the time so the report can be inherited/manipulated before output from clients.

lib/spoom/cli/commands/coverage.rb
```diff
@@ -103,6 +103,47 @@ def timeline
...
+          files = Dir.glob("#{data_dir}/*.json")
+          if files.empty?
+            message_no_data(data_dir)
+            exit(1)
+          end
+
+          snapshots = files.sort.map do |file|
+            json = File.read(file)
+            Spoom::Coverage::Snapshot.from_json(json)
+          end.filter(&:commit_timestamp).sort_by!(&:commit_timestamp)
```
KaanOzkan, line 122: We actually have items/commits without a timestamp here? 😲

lib/spoom/config.rb
```diff
@@ -6,5 +6,6 @@ module Config
     SORBET_CONFIG = "sorbet/config"
     SORBET_GEM_PATH = Gem::Specification.find_by_name("sorbet-static").full_gem_path
     SORBET_PATH = (Pathname.new(SORBET_GEM_PATH) / "libexec" / "sorbet").to_s
+    SPOOM_PATH = (Pathname.new(__FILE__) / ".." / ".." / "..").to_s
```
KaanOzkan, line 9: I'm curious what's going on here. I think you are moving back to the root directory of this repo but i don't know how. 

What does the `/` do?

lib/spoom/coverage.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    sig { params(path: String).returns(Snapshot) }
+    def self.snapshot(path: '.')
+      snapshot = Snapshot.new
+      metrics = Spoom::Sorbet.srb_metrics(path: path, capture_err: true)
+      return snapshot unless metrics
+
+      sha = Spoom::Git.last_commit(path: path)
+      snapshot.commit_sha = sha
+      snapshot.commit_timestamp = Spoom::Git.commit_timestamp(sha, path: path).to_i if sha
```
KaanOzkan, line 22: I guess my previous comment is invalid now and the question is this `sha` can be null 😲

KaanOzkan (approved):
I skimmed over the CSS/D3JS parts 😕

lib/spoom/config.rb
paracycle: `Pathname#/` is an alias to `Pathname#join` :)

paracycle (approved):
Some comments about removing some of the presentational things from the Template classes and moving them into the actual templates, so that the template is just build from the data in the Template object itself, no extra presentational stuff coming from Ruby code.

Otherwise, this looks great.

lib/spoom/cli/commands/coverage.rb
Morriar: Spoom allows you to take a snapshot of a project that doesn't uses Git so no commit sha nor timestamp (and also no timeline).

Morriar:
> Some comments about removing some of the presentational things from the Template classes and moving them into the actual templates, so that the template is just built from the data in the Template object itself, no extra presentational stuff coming from Ruby code.

I've been doing back and forth on that.

Basically what I wanted (for example in Sorbet-Metrics) was to be able to customize the header/footer and style while controlling the content (through the cards) but without rewriting the template files.

The rationale behind this version is that you can customize the default template directly from your script without having to duplicate it thanks to inheritance.

I'm happy to change it but I feel the proposed approach would force me to rewrite the whole template file just to change a few details in each client.

paracycle:
> The rationale behind this version is that you can customize the default template directly from your script without having to duplicate it thanks to inheritance.

OK, that's fair and is a legitimate use-case.

> I'm happy to change it but I feel the proposed approach would force me to rewrite the whole template file just to change a few details in each client.

I don't feel that strongly about it but I feel like, ultimately, the best thing is to customize templates instead of code. But we can evolve that as part of later work, not a requirement for this PR to be shipped as is.

PR #39, opened 2020-10-13
Add Snapshot metrics for singleton classes

Morriar:
Since https://github.com/sorbet/sorbet/pull/3446 was merged upstream we can provide a more accurate number for the amount of classes in the project by subtracting the number of singleton classes.

This PR changes the output of `spoom coverage snapshot` and the related tests to reflect that.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -45,10 +45,10 @@ GEM
     rubocop-sorbet (0.4.0)
       rubocop
     ruby-progressbar (1.10.1)
-    sorbet (0.5.5916)
-      sorbet-static (= 0.5.5916)
-    sorbet-runtime (0.5.5916)
-    sorbet-static (0.5.5916-universal-darwin-14)
+    sorbet (0.5.5943)
+      sorbet-static (= 0.5.5943)
+    sorbet-runtime (0.5.5943)
+    sorbet-static (0.5.5943-universal-darwin-14)
     thor (1.0.1)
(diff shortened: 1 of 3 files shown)


paracycle (commented):
What happens if the user is using this with an older `sorbet-static` version? This PR only bumps the version of it this gem uses for development.

Morriar:
> What happens if the user is using this with an older `sorbet-static` version? This PR only bumps the version of it this gem uses for development.

We never directly show that metrics (it's not super useful for the end user), we just substract its value from the `classes` one that never changed. What really changed in the output if that I removed the `(including singleton classes)` warning.

So the difference is that if you run with a version of sorbet < 0.5.5939, you will have more classes since singleton_classes will be initialized by default at 0.

> I think it would be great to do a version check for Sorbet and output something like We can't print singleton class metrics on versions of sorbet-static < 0.5.5xxx. That would allow users to know how to get better output.

It's a bit more complex because the snapshot itself doesn't know everything about the real version ran. If you run `spoom` with `--path` and without `--bundle-install` the version in the snapshot (read from the Gemfile) might not be the real version we collected the metrics (the one ran by `spoom`).

tbh, I'm not super inclined in adding more complexity for that metrics since we don't really track it anywhere at the moment and both `sorbet` and `spoom` are in beta versions, so this could change at any time.

paracycle (approved):
Apologies, I seem to have missed the reply last week. All good!

PR #40, opened 2020-10-20
Avoid crashing on non-existing path when calling `git`

Morriar:
There is no way to reach this state with normal CLI use but it's possible to get to this state using the API directly.

Since all the methods in this module return an error string + status if something go wrong with the `git` command, we reuse the same mechanism instead of raising for this error so client only have one thing to check.

This change also makes testing with incomplete file hierarchies easier.

### DIFF
--- a/lib/spoom/git.rb
+++ b/lib/spoom/git.rb
@@ -11,6 +11,7 @@ module Spoom
     # Execute a `command`
     sig { params(command: String, arg: String, path: String).returns([String, String, T::Boolean]) }
     def self.exec(command, *arg, path: '.')
+      return "", "Error: `#{path}` is not a directory.", false unless File.directory?(path)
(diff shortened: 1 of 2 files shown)


paracycle (approved):
Lol, what is this business with returning Error objects everywhere? Go? 😂

KaanOzkan (approved):
>Lol, what is this business with returning Error objects everywhere? Go? 😂

I like it 😄

Morriar:
Agreed, I could replace this clean exceptions.

Or by Adt::Result or the horrendous [MaybeError](https://github.com/nitlang/nit/blob/74398ebe85ee2337efa986205531b3ca696c8543/lib/core/error.nit#L42-L91) wrapper we have in Nit until we figure out how to implement exceptions properly 😂

PR #43, opened 2020-11-17
Allow to customize colors

Morriar:
Add a bunch of options so the user can configure the colors used for each strictness in the visualizations:

```bash
$ spoom coverage report \
  --color-true "#648ffe" \
  --color-false "#fe6002" \
  --color-ignore "#feb000" \
  --color-strict "#795ef0" \
  --color-strong "#6444f1"
```

Will give the following output:

image
image
image

Helpful for colorblind folks.

### DIFF
--- a/lib/spoom/coverage/d3.rb
+++ b/lib/spoom/coverage/d3.rb
@@ -10,6 +10,12 @@ module Spoom
     module D3
       extend T::Sig
 
+      COLOR_IGNORE = "#999"
+      COLOR_FALSE = "#db4437"
+      COLOR_TRUE = "#0f9d58"
+      COLOR_STRICT = "#0a7340"
+      COLOR_STRONG = "#064828"
+
       sig { returns(String) }
       def self.header_style
         <<~CSS
@@ -50,23 +56,25 @@ module Spoom
         CSS
       end
 
-      sig { returns(String) }
-      def self.header_script
+      sig { params(palette: ColorPalette).returns(String) }
+      def self.header_script(palette)
         <<~JS
           var parseDate = d3.timeParse("%s");
 
           function strictnessColor(strictness) {
             switch(strictness) {
+              case "ignore":
+                return "#{palette.ignore}";
               case "false":
-                return "#db4437";
+                return "#{palette.false}";
               case "true":
-                return "#0f9d58";
+                return "#{palette.true}";
               case "strict":
-                return "#0a7340";
+                return "#{palette.strict}";
               case "strong":
-                return "#064828";
+                return "#{palette.strong}";
             }
-            return "#999";
+            return "#{palette.false}";
           }
 
           function toPercent(value, sum) {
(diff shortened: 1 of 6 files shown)


lib/spoom/coverage/d3.rb
```diff
@@ -50,23 +56,25 @@ def self.header_style
...
               case "strict":
-                return "#0a7340";
+                return "#{palette.strict}";
               case "strong":
-                return "#064828";
+                return "#{palette.strong}";
             }
-            return "#999";
+            return "#{palette.false}";
           }
```
paracycle, line 78: What if this was a hash lookup:
```js
const COLORS = {
  ignore: (palette) => palette.ignore,
  false: (palette) => palette.false,
  true: (palette) => palette.true,
  strict: (palette) => palette.strict,
  strong: (palette) => palette.strong,
}

const DEFAULT_COLOR = (palette) => palette.false;

function strictnessColor(strictness) {
  let strictnessColorLookup = COLORS[strictness] || DEFAULT_COLOR;
  return strictnessColorLookup(palette);
}
```
or even:
```js
const STRICTNESS_LEVELS = ["ignore", "false", "true", "strict", "strong"];

function strictnessColor(strictness) {
  strictness = STRICTNESS_LEVELS.includes(strictness) ? strictness : "false";
  return palette[strictness];
}
```

paracycle (approved):

lib/spoom/coverage/d3.rb
paracycle: Oh, I just realized that `palette` is not even a JS variable. Scratch what I said above.

Morriar: It's called a crazy amount of time when generating the circle map for big apps (ie. our core), won't the hash lookup slow the rendering too much?

paracycle: AFAIK, JS is optimized for hash lookups, since all objects are hashes and all that.

This might work much faster, but I am not sure:
```js
const COLORS = {
  ignore: "#{palette.ignore}",
  false: "#{palette.false}",
  true: "#{palette.true}",
  strict: "#{palette.strict}",
  strong: "#{palette.strong}",
}

const DEFAULT_COLOR = "#{palette.false}";

function strictnessColor(strictness) {
  return COLORS[strictness] || DEFAULT_COLOR;
}
```

Morriar: I was curious so I ran a few benches.

I used three versions, one using the hash (as you proposed), one using a switch (as already used) and one used an if (for completeness).

Here's the benchmarking code (`sum` is only used to make sure the processor does not skip instructions seeing no side effect):

```js
const items = ["ignore", "true", "false", "strict", "strong"]
var sum = 0

for (var i = 0; i < 100000000; i++) {
  var max = items.length;
  var item = items[Math.floor(Math.random() * max)]
  sum += strictnessColor(item);
}

console.log("done: " + sum)
```

And here are the results (in seconds) for 5 runs:


Runs | Hash | Switch | If
-- | -- | -- | --
1 | 3.353 | 2.931 | 2.898
2 | 3.484 | 2.948 | 2.928
3 | 3.399 | 2.931 | 2.895
4 | 3.409 | 2.922 | 2.921
5 | 3.429 | 2.94 | 2.957
**AVG** | **3.4148** | **2.9344** | **2.9198**

So the switch and if implementations are a bit faster. (I _think_ they are actually producing the same low level instructions)

I'm super surprised the hash is that fast to be honest. TIL!
My guess for the overhead is the hashing of the key.

PR #44, opened 2020-11-17
Update ruby setup version for Github workflow

Morriar:
So we can fix this error on CI: https://github.com/Shopify/spoom/pull/43/checks?check_run_id=1412893423

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -24,7 +24,7 @@ jobs:
     # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
     # change this to (see https://github.com/ruby/setup-ruby#versioning):
     # uses: ruby/setup-ruby@v1
-      uses: ruby/setup-ruby@ec106b438a1ff6ff109590de34ddc62c540232e0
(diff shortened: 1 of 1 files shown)


.github/workflows/ruby.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     # To automatically get bug fixes and new Ruby versions for ruby/setup-ruby,
     # change this to (see https://github.com/ruby/setup-ruby#versioning):
     # uses: ruby/setup-ruby@v1
-      uses: ruby/setup-ruby@ec106b438a1ff6ff109590de34ddc62c540232e0
```
paracycle, line 27: I just use `ruby/setup-ruby@v1` like it says in the comment above to get automatic features and bug fixes. We can do the same instead of chasing versions.

paracycle (approved):

.github/workflows/ruby.yml
```diff
@@ -24,7 +24,7 @@ jobs:
...
+      uses: ruby/setup-ruby@34fbbc1d36fbeb08e98a6bdb591cac5eed7effa8
```
Morriar, line 27:
```suggestion
      uses: ruby/setup-ruby@v1
```

Issue #48, opened 2020-11-22
Error: Uninitialized constant Spoom::Config::Pathname


nvh0412:
# Description
Tried running Spoom to generate Sorbet report but I got `uninitialized constant Spoom::Config::Pathname` error. Seems like we forgot to require 'pathname' for `lib/spoom/config.rb`

- Ruby 2.6.5
- Sinatra 2.0.5
- Sequel 4.45.0

Morriar:
👋 Hey @nvh0412,

Can you provide the steps and commands you ran to reproduce this error?

Thanks!

nvh0412:
Hi @Morriar 

I just created another ruby project (Ruby 2.6.6) with a simple Gemfile which includes spoom@1.0.6

![Screen Shot 2020-11-24 at 19 34 34](https://user-images.githubusercontent.com/1973623/100094921-53227b80-2e8c-11eb-97b6-c4ca959a5ef0.png)

nvh0412:
@Morriar  I think you forgot to use `spoom` with a non-rails project. If we use spoom with Rails project, we don't get this error because[ rails engine already include `pathname` for us already.](https://github.com/rails/rails/blob/fb3ab475ad985f5683f9b7768e16e5d89da93739/railties/lib/rails.rb#L5)

That's why I created a PR to fix this issue https://github.com/Shopify/spoom/pull/49

Morriar:
@nvh0412, thanks for the repro steps 🙏 

We generally use `spoom` through `bundle exec` so you might find more bugs down the line.

connorshea:
Could we get a new release with this fix? I just ran into the same thing :)

Morriar:
@connorshea I just released 1.0.7: https://rubygems.org/gems/spoom/versions/1.0.7

connorshea:
Thsnks!

PR #49, opened 2020-11-22
require pathname to fix Unitialized constant

nvh0412:
# Description

Resolve https://github.com/Shopify/spoom/issues/48

# How to test

- [x] Try applying this patch with my local project, it is working

### DIFF
--- a/lib/spoom/config.rb
+++ b/lib/spoom/config.rb
@@ -1,5 +1,6 @@
 # typed: true
 # frozen_string_literal: true
+
 require 'pathname'
 
 module Spoom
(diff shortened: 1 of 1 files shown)


lib/spoom/config.rb
```diff
@@ -1,5 +1,6 @@
 # typed: true
 # frozen_string_literal: true
+require 'pathname'
```
Morriar, line 3:
```suggestion

require 'pathname'
```

Morriar (approved):
Thanks 🙏

PR #52, opened 2020-11-27
Fixes encoding issues

masylum:
While running certain commands, spoom was chocking with some of our files containing € signs.

### DIFF
--- a/lib/spoom/sorbet/sigils.rb
+++ b/lib/spoom/sorbet/sigils.rb
@@ -56,14 +56,14 @@ module Spoom
       sig { params(path: T.any(String, Pathname)).returns(T.nilable(String)) }
       def self.file_strictness(path)
         return nil unless File.exist?(path)
-        content = File.read(path)
+        content = File.read(path, encoding: 'UTF-8')
         strictness_in_content(content)
       end
 
       # changes the sigil in the file at the passed path to the specified new strictness
       sig { params(path: T.any(String, Pathname), new_strictness: String).returns(T::Boolean) }
       def self.change_sigil_in_file(path, new_strictness)
-        content = File.read(path)
+        content = File.read(path, encoding: 'UTF-8')
         new_content = update_sigil(content, new_strictness)
 
         File.write(path, new_content)
--- a/test/spoom/sorbet/sigils_test.rb
+++ b/test/spoom/sorbet/sigils_test.rb
@@ -74,6 +74,16 @@ module Spoom
         assert_equal("no", strictness)
       end
 
+      def test_file_with_non_utf8_encoding
+        content = <<~STR
+          # typed: true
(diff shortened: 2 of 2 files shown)


Morriar (changes requested):
👋 @masylum, 

Good catch, thanks a lot for the fix!

Can you add tests in https://github.com/Shopify/spoom/blob/master/test/spoom/sorbet/sigils_test.rb showing the behaviours for `file_strictness` and `change_sigil_in_file` please?

masylum:
Hey @Morriar, any feedback?

test/spoom/sorbet/sigils_test.rb
```diff
@@ -74,6 +74,16 @@ class A; end
         assert_equal("no", strictness)
       end
 
+      def test_file_with_non_utf8_encoding
+        content = <<~STR
+          # typed: true
+          "#{[0x89].pack('c*')}"
+        STR
+
+        strictness = Sigils.strictness_in_content(content.dup.force_encoding("ISO-8859-1"))
```
Morriar, line 83: Why not something like `"€".encode("ISO-8859-15")` so it's easier to read?

test/spoom/sorbet/sigils_test.rb
```diff
@@ -74,6 +74,16 @@ class A; end
         assert_equal("no", strictness)
       end
 
+      def test_file_with_non_utf8_encoding
+        content = <<~STR
+          # typed: true
+          "#{[0x89].pack('c*')}"
```
Morriar, line 80: Can we have a comment indicating which character this is supposed to represent?

lib/spoom/sorbet/sigils.rb
```diff
@@ -56,14 +56,14 @@ def self.update_sigil(content, new_strictness)
       sig { params(path: T.any(String, Pathname)).returns(T.nilable(String)) }
       def self.file_strictness(path)
         return nil unless File.exist?(path)
-        content = File.read(path)
+        content = File.read(path, encoding: 'UTF-8')
```
Morriar, line 59: Will this solve the problem? We're still reading an ISO file as UTF8?

Furthermore, would that mean we'll save the file as UTF after the bump? Silently re-encoding users files might not be the best approach here.

Morriar (changes requested):
👋 @masylum,

I'm not sure this is the right approach.

This still reads the file with the wrong encoding, then saves it as UTF8 once the sigil bump is done. This might create issues for the user where we'll silently recode their files.

Also, the provided test doesn't use external files, so it doesn't really reproduce the original issue you were having.

I proposed another fix for this issue in https://github.com/Shopify/spoom/pull/56, let me know if it works for you. In the mean time, I'm closing this PR.

PR #56, opened 2020-12-16
Allow bumping files with non-utf8 content

Morriar:
As @masylum found out, when trying to get the sigil from a file, we're reading it with the default encoding: `utf-8`.

This is problematic when reading a file from a different encoding, especially when the file contains special characters like `À coûté 10€`.

This pull-request allows the user to read and write the files with a specific encoding by adding:
* a `encoding:` keyword argument to `Sigils` services
* a `--encoding` option to the `bump` command

Thus, solving the issue raised by #52.

### DIFF
--- a/lib/spoom/sorbet/sigils.rb
+++ b/lib/spoom/sorbet/sigils.rb
@@ -87,7 +87,7 @@ module Spoom
           extension: String
         ).returns(T::Array[String])
       end
-      def self.files_with_sigil_strictness(directory, strictness, extension = ".rb")
+      def self.files_with_sigil_strictness(directory, strictness, extension: ".rb")
         paths = Dir.glob("#{File.expand_path(directory)}/**/*#{extension}").sort.uniq
         paths.filter do |path|
           file_strictness(path) == strictness
--- a/lib/spoom/sorbet/sigils.rb
+++ b/lib/spoom/sorbet/sigils.rb
@@ -56,14 +56,14 @@ module Spoom
       sig { params(path: T.any(String, Pathname)).returns(T.nilable(String)) }
       def self.file_strictness(path)
         return nil unless File.exist?(path)
-        content = File.read(path)
+        content = File.read(path, encoding: Encoding::ASCII_8BIT)
         strictness_in_content(content)
       end
 
       # changes the sigil in the file at the passed path to the specified new strictness
       sig { params(path: T.any(String, Pathname), new_strictness: String).returns(T::Boolean) }
       def self.change_sigil_in_file(path, new_strictness)
-        content = File.read(path)
+        content = File.read(path, encoding: Encoding::ASCII_8BIT)
         new_content = update_sigil(content, new_strictness)
 
         File.write(path, new_content)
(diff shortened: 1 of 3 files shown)


KaanOzkan (approved):

lib/spoom/cli/bump.rb
```diff
@@ -19,13 +19,15 @@ class Bump < Thor
         desc: "Change files to this strictness"
       option :force, type: :boolean, default: false, aliases: :f,
         desc: "Change strictness without type checking"
+      option :encoding, type: :string, desc: "Use specific encoding when reading and writing files"
```
paracycle, line 22: Can we not always read/write files using `Encoding::ASCII_8BIT`, which basically means raw binary?

This seems to work:

```ruby
# Setup the file
str = 'puts "À coûté 10€"'
# => "puts \"À coûté 10€\""
File.write("foo.rb", str.encode("ISO-8859-15"))

# Actual Spoom routine will look like this
new_str = File.read("foo.rb", encoding: Encoding::ASCII_8BIT)
new_str = "typed: true\n" + new_str
File.write("foo.rb", new_str)

# Assert file contents
File.read("foo.rb", encoding: "iso-8859-15" )
# => "typed: true\nputs \"\xC0 co\xFBt\xE9 10\xA4\""
File.read("foo.rb", encoding: "iso-8859-15" ).encode("UTF-8")
# => "typed: true\nputs \"À coûté 10€\""
```

masylum (commented):
That's much nicer than my PR 😅  Awesome!

lib/spoom/cli/bump.rb
Morriar: ^ This is pure genius! ❤️

lib/spoom/sorbet/sigils.rb
```diff
@@ -56,14 +56,14 @@ def self.update_sigil(content, new_strictness)
       sig { params(path: T.any(String, Pathname)).returns(T.nilable(String)) }
       def self.file_strictness(path)
         return nil unless File.exist?(path)
-        content = File.read(path)
+        content = File.read(path, encoding: Encoding::ASCII_8BIT.to_s)
```
paracycle, line 59: Super nitpicky but you really should not need the `.to_s` here.

lib/spoom/sorbet/sigils.rb
```diff
@@ -56,14 +56,14 @@ def self.update_sigil(content, new_strictness)
...
         strictness_in_content(content)
       end
 
       # changes the sigil in the file at the passed path to the specified new strictness
       sig { params(path: T.any(String, Pathname), new_strictness: String).returns(T::Boolean) }
       def self.change_sigil_in_file(path, new_strictness)
-        content = File.read(path)
+        content = File.read(path, encoding: Encoding::ASCII_8BIT.to_s)
```
paracycle, line 66: Same as above.

paracycle (approved):

lib/spoom/sorbet/sigils.rb
Morriar: Good call: https://github.com/sorbet/sorbet/pull/3861

Morriar: Fixed 🎉

PR #59, opened 2021-01-04
Do not include RBI by default in coverage metrics

Morriar:
By default, running `spoom coverage` will output the coverage metrics **including** the RBi files.

This can give misleading metrics when having a large amount of RBIs (for example when they are generated by `tapicoa`).

This PR makes the default to not include RBIs and adds the `--include-rbi` option to count them:

Example for `spoom coverage` on `spoom` itself:

```
Content:
  files: 68
  modules: 52
  classes: 154
  methods: 1994

Sigils:
  true: 49 (72%)
  strict: 19 (28%)

Methods:
  with signature: 300 (15%)
  without signature: 1694 (85%)

Calls:
  typed: 4893 (80%)
  untyped: 1246 (20%)
```

Example for `spoom coverage --include-rbi`:

```
Content:
  files: 49
  modules: 16
  classes: 79
  methods: 687

Sigils:
  true: 30 (61%)
  strict: 19 (39%)

Methods:
  with signature: 300 (44%)
  without signature: 387 (56%)

Calls:
  typed: 4861 (79%)
  untyped: 1278 (21%)
```

Also added a bunch of tests.

### DIFF
--- a/lib/spoom/coverage.rb
+++ b/lib/spoom/coverage.rb
@@ -59,11 +59,14 @@ module Spoom
       )
     end
 
+    sig { params(path: String).returns(Sorbet::Config) }
+    def self.sorbet_config(path: ".")
+      Sorbet::Config.parse_file("#{path}/#{Spoom::Config::SORBET_CONFIG}")
+    end
+
     sig { params(path: String).returns(FileTree) }
     def self.sigils_tree(path: ".")
-      config_file = "#{path}/#{Spoom::Config::SORBET_CONFIG}"
-      return FileTree.new unless File.exist?(config_file)
-      config = Sorbet::Config.parse_file(config_file)
+      config = sorbet_config(path: path)
       files = Sorbet.srb_files(config, path: path)
       files.select! { |file| file =~ /\.rb$/ }
       files.reject! { |file| file =~ %r{/test/} }
--- a/lib/spoom/coverage.rb
+++ b/lib/spoom/coverage.rb
@@ -11,10 +11,17 @@ module Spoom
   module Coverage
     extend T::Sig
 
-    sig { params(path: String).returns(Snapshot) }
-    def self.snapshot(path: '.')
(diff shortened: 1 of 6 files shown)


KaanOzkan (approved):

lib/spoom/coverage.rb
```diff
@@ -11,10 +11,19 @@ module Spoom
...
-    def self.snapshot(path: '.')
+    sig { params(path: String, rbi: T::Boolean).returns(Snapshot) }
+    def self.snapshot(path: '.', rbi: false)
+      config = sorbet_config(path: path)
+      config.allowed_extensions.push(".rb", ".rbi") if config.allowed_extensions.empty?
+
+      new_config = Sorbet::Config.new
+      new_config.paths.concat(config.paths)
+      new_config.ignore.concat(config.ignore)
+      new_config.allowed_extensions.concat(config.allowed_extensions.reject { |ext| !rbi && ext == ".rbi" })
```
paracycle, line 22: It might be a bit hacky but a quick and dirty way of deep cloning `Config` in an encapsulated way would be to add the method:
```
def deep_clone
  Marshal.load(Marshal.dump(self))
end
```
to the `Config` class.

With that, this code becomes:
```suggestion
      new_config = config.deep_clone
      new_config.allowed_extensions.concat(config.allowed_extensions.reject { |ext| !rbi && ext == ".rbi" })
```

paracycle (approved):
Looks good to me. As we discussed offline, ignoring RBI files completely might give us the wrong "sigged calls" metric, but this is a good first step.

lib/spoom/coverage.rb
Morriar: This looks less good when you add all the casts to make it statically valid.

I factorized away the copy of the object in 005dc93.

Morriar:
Finally I kept the default behaviour as calls are our most important target metric.

I changed the option name to `spoom coverage --no-rbi` so the user can still access the accurate data.

PR #63, opened 2021-01-07
Fix an assert that should have been an assert_equal

Morriar:
Proof we need sigs on Minitest.


paracycle (approved):
I guess it was matching the second argument to `msg`. Good catch!

KaanOzkan (approved):

PR #64, opened 2021-01-13
Quality of life improvements to spoom bump

Morriar:
This pull-request adds a few options to `spoom bump` as to make it easier to use in more complex bumps.

1. Add some printing so we see what the command is doing
2. Add `spoom bump --dry to not do anything, just see the changes that might happen
3. Add `spoom bump --only file.rb` to change only the files listed
4. Adjust exit values so `spoom bump` can more easily used in a script

We can also use all of this to check if files could be bumped to `typed: strict` in our own CI.

See the pull-request #65 for example:

image

As a matter of fact, this PR is also enabling `spoom bump--from true --to strict --dry` on this repository.

Closes #65.

### DIFF
--- a/lib/spoom/cli/bump.rb
+++ b/lib/spoom/cli/bump.rb
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: true
 # frozen_string_literal: true
 
 require 'find'
@@ -27,6 +27,7 @@ module Spoom
         from = options[:from]
         to = options[:to]
         force = options[:force]
+        exec_path = File.expand_path(self.exec_path)
 
         unless Sorbet::Sigils.valid_strictness?(from)
           say_error("Invalid strictness #{from} for option --from")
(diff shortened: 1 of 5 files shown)


KaanOzkan (approved):
I really like `--dry` and its potential usages in CI ❤️ 

This is a general comment about `spoom bump` but this usage got me thinking. In the future clients might want to bump every file to highest strictness possible.

Since the beginning `spoom bump` works with specific `--from` and `--to` values and this PR is again relying on them and increasing that coupling.

I just wanted to point this out. I don't see a reason to change things now but the above usage/expectation seems very plausible to me, especially in terms of CI.

Maybe if we do support it we don't want to change spoom internals at all and instead run each permutation of bumps with a single command and report to user 😄 🤷‍♀️

lib/spoom/cli/bump.rb
```diff
@@ -57,7 +83,35 @@ def bump(directory = ".")
...
+        undo_changes(files_with_errors, from)
+
+        files_changed = files_to_bump - files_with_errors
+        print_changes(files_changed, from: from, to: to, dry: dry, path: exec_path)
+        undo_changes(files_to_bump, from) if dry
+        exit(files_changed.empty?)
+      end
+
+      no_commands do
+        def print_changes(files, from: "false", to: "true", dry: false, path: File.expand_path("."))
```
RyanBrushett, line 95: Is it possible to add a sig to this method and `undo_changes` to get this file back to `typed: strict`? Or does the `no_commands` block cause some hiccup here?

RyanBrushett (approved):
This is great!

Morriar:
Thanks for you review guys!

@KaanOzkan:

> This is a general comment about `spoom bump` but this usage got me thinking. In the future clients might want to bump every file to highest strictness possible.

In that case you might want to use `sorbet --suggest-typed`. The idea of `bump` was to allow us to go by step and avoid adding `strict` in our codebase so we don't force future users to add signatures everywhere.

> Since the beginning `spoom bump` works with specific `--from` and `--to` values and this PR is again relying on them and increasing that coupling.

That being said, we can always change the options to be `--from fase,true --to strict`. Or decide that no option means `any`.

> This feature seems big enough to be mentioned in readme under `bump` section?

Good catch, I added a few lines about it.

lib/spoom/cli/bump.rb
Morriar: It is possible for Sorbet but Thor will complain at runtime about your `no_commands` bein a command and not having a documentation. We need to take a look at this at some point.

One solution is to move those helper method to a real helper like it's done here: https://github.com/Shopify/spoom/blob/master/lib/spoom/cli/helper.rb

You can observe the impact of that limitation on the strictness of spoom's cli modules:

image

(the only one at `typed: strict` is the helper)

PR #69, opened 2021-02-09
Improvements to `spoom bump`

Morriar:
This pull-request does two things:

1. It adds a `--suggest-bump-command` option to the bump command so one can specify how the bump should be ran:

   ```bash
   $ bundle exec spoom bump --dry -f --suggest-bump-command "bundle exec spoom bump"
   Can bump 1 file from false to true:
    + file1.rb

   Run `bundle exec spoom bump --from false --to true` to bump them
   ```

2. It adds a line when the command starts so the user can see something is happening:

   ```bash
   $ bundle exec spoom bump
   Checking files...
   ```

### DIFF
--- a/lib/spoom/cli/bump.rb
+++ b/lib/spoom/cli/bump.rb
@@ -24,6 +24,8 @@ module Spoom
         desc: "Only display what would happen, do not actually change sigils"
       option :only, type: :string, default: nil, aliases: :o,
         desc: "Only change specified list (one file by line)"
+      option :suggest_bump_command, type: :string, default: "spoom bump",
+        desc: "Command to suggest if files can be bumped"
       sig { params(directory: String).void }
       def bump(directory = ".")
         in_sorbet_project!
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
It's unintuitive to me that the option is named `suggest-bump-command` but we append `--from & --to true` to the command user passed. But it makes sense why we do that so I guess it's okay.

Morriar:
> It's unintuitive to me that the option is named `suggest-bump-command` but we append `--from & --to true` to the command user passed. But it makes sense why we do that so I guess it's okay.

Good point. Would it be better if we didn't happen anything and let the user provide the whole command? So if you use a script for example you can directly point to it?

KaanOzkan:
If we trust the user yes. I don't mind that it's being safe right now. Without it, you could suggest a wrong command by accident, ie. suggest `--from false --to strict` when the spoom bump that was executed actually used `--from false --to true`. 

I wanted to point out that it's unintuitive but it makes sense if we want to be safe 🤷

lucianosousa (approved):
Thanks for doing that. Truly appreciate

Morriar:
@KaanOzkan I changed it so the user has to provide the command options too 👍

KaanOzkan (approved):

PR #70, opened 2021-02-10
Testing against multiple ruby versions

lucianosousa:
This PR aims to improve the GH action to test against multiple ruby versions.
Right now we only know that it works with Ruby 2.6

I'm also introducing tests against Ruby 3.0. It's going to be easier to keep compatibility with all of the stable ruby versions we have right now

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,5 @@ gemspec
 group(:development) do
   gem('rubocop-shopify', require: false)
   gem('rubocop-sorbet', require: false)
-  gem('byebug')
   gem('pry-byebug')
 end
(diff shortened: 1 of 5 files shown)


Gemfile
```diff
@@ -8,6 +8,5 @@ gemspec
 group(:development) do
   gem('rubocop-shopify', require: false)
   gem('rubocop-sorbet', require: false)
-  gem('byebug')
```
lucianosousa, line 11: This is not needed here because `byebug` is a dependency of `pry-byebug`

Morriar (approved):
Amazing, thanks!

KaanOzkan (approved):
Thanks

PR #71, opened 2021-02-10
Caching gems during GH actions

lucianosousa:
Adding the cache flag.

This is currently failing because of an issue with Sorbet in the repo.

@Morriar is aware and we can merge it once this is fixed


Morriar:
Superseded by https://github.com/Shopify/spoom/pull/73.

PR #75, opened 2021-03-09
Condenses default `spoom tc` output and makes the sort option default

seachel:
This PR makes the default output of the `spoom tc` command to be the condensed form of error messages displayed sorted.

Current errors presentation running `spoom tc`:

image

Proposed error presentation running `spoom tc`:

image


Morriar (approved):
Thanks!

PR #76, opened 2021-03-15
Add option to count errors if files were bumped to true

vinistock:
Add `--count-errors` option to bump, which is similar to a dry run, but prints the number of total errors if all files were bumped to true. No changes are made to the files.

This is useful for measuring typing progress in big code bases where addressing a specific signature or error doesn't unlock the file being `typed: true`, but already reduces the remaining errors to fix.

This is how the output looks. (`bspoom` is just an alias for `bin/spoom`).
Screen Shot 2021-03-15 at 3 36 09 PM

### DIFF
--- a/lib/spoom/cli/bump.rb
+++ b/lib/spoom/cli/bump.rb
@@ -26,6 +26,8 @@ module Spoom
         desc: "Only change specified list (one file by line)"
       option :suggest_bump_command, type: :string,
         desc: "Command to suggest if files can be bumped"
+      option :count_errors, type: :boolean, default: false,
+        desc: "Count the number of errors if all files were bumped"
       sig { params(directory: String).void }
       def bump(directory = ".")
         in_sorbet_project!
@@ -45,6 +47,11 @@ module Spoom
 
         unless Sorbet::Sigils.valid_strictness?(to)
           say_error("Invalid strictness `#{to}` for option `--to`")
+          exit(1)
+        end
+
+        if options[:count_errors] && !dry
+          say_error("`--count-errors` can only be used with `--dry`")
           exit(1)
         end
 
@@ -94,6 +101,8 @@ module Spoom
         end.compact.uniq
 
         undo_changes(files_with_errors, from)
+
+        say("Found #{errors.length} type checking error#{'s' if errors.length > 1}") if options[:count_errors]
 
         files_changed = files_to_bump - files_with_errors
         print_changes(files_changed, command: cmd, from: from, to: to, dry: dry, path: exec_path)
--- a/test/spoom/cli/bump_test.rb
+++ b/test/spoom/cli/bump_test.rb
@@ -367,6 +367,48 @@ module Spoom
         assert_equal("true", Sorbet::Sigils.file_strictness("#{@project.path}/file1.rb"))
         assert_equal("false", Sorbet::Sigils.file_strictness("#{@project.path}/vendor/file2.rb"))
       end
+
+      def test_count_errors_without_dry
+        @project.write("file1.rb", <<~RB)
+          # typed: false
+          class Foo
+            def foo
+            end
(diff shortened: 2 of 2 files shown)


lib/spoom/cli/bump.rb
```diff
@@ -95,9 +97,13 @@ def bump(directory = ".")
...
         undo_changes(files_with_errors, from)
 
+        if options[:count_errors]
+          say("Errors: #{errors.length}")
+        end
+
         files_changed = files_to_bump - files_with_errors
         print_changes(files_changed, command: cmd, from: from, to: to, dry: dry, path: exec_path)
-        undo_changes(files_to_bump, from) if dry
+        undo_changes(files_to_bump, from) if dry || options[:count_errors]
```
Morriar, line 106: Why do we need this?

lib/spoom/cli/bump.rb
```diff
@@ -95,9 +97,13 @@ def bump(directory = ".")
 
         undo_changes(files_with_errors, from)
 
+        if options[:count_errors]
+          say("Errors: #{errors.length}")
+        end
```
Morriar, line 102:
```suggestion
        say("Errors: #{errors.length}") if options[:count_errors]
```

test/spoom/cli/bump_test.rb
```diff
@@ -367,6 +367,29 @@ class A; end
...
+
+          Foo.new.foos
+        RB
+
+        out, err, status = @project.bundle_exec("spoom bump --no-color --count-errors")
+        assert_empty(err)
+        assert_equal(<<~OUT, out)
+          Checking files...
+
+          Errors: 1
```
Morriar, line 387: Reading this it looks like the errors are about the `bump` itself.

What do you think about something like "Found X type checking error(s)."?

lib/spoom/cli/bump.rb
vinistock: In my opinion, it would be surprising for this option to also bump the files. At least I'd expect `--count-errors` to just count errors and not attempt to keep the files bumped.

Do you feel like the other way around is more intuitive?

Morriar: In that case, can we make it so you have to pass both `dry` and `count-errors`?

Like 

```
$ spoom bump --count-errors
Error: `--count-errors` can only be used with `--dry`.
```

?

test/spoom/cli/bump_test.rb
```diff
@@ -367,6 +367,29 @@ class A; end
...
+
+          Foo.new.foos
+        RB
+
+        out, err, status = @project.bundle_exec("spoom bump --no-color --count-errors")
+        assert_empty(err)
+        assert_equal(<<~OUT, out)
+          Checking files...
+
+          Found 1 type checking error(s).
```
Morriar, line 387:
```suggestion
          Found 1 type checking error(s)
```

We never use `.` at the end of the message in `spoom` (you can open another PR if you want to change them all 😂)

Also I think there is already some logic round the `s` plural form for `X file(s?) bumped`, let's do the same for `error(s?)` so the messages are uniform please 🙏

lib/spoom/cli/bump.rb
vinistock: Yeah, that makes more sense. 👍

test/spoom/cli/bump_test.rb
vinistock: Done!

Morriar (approved):
Thanks for indulging me! 🙏

RyanBrushett (approved):

Morriar:
@vinistock I just realized we should also add an entry in the README about this option: https://github.com/Shopify/spoom/blob/master/README.md#change-the-sigil-used-in-files

PR #78, opened 2021-03-18
Fallback to older srb if necessary

vinistock:
Instead of making our Sorbet version requirement more strict, we should determine whether to use the new `--no-error-sections` option based on the executable version.

This PR checks the version and decides whether to use the argument or not. On the second commit, I also mentioned the new `--count-errors` option in the README, which we missed in #76.

### DIFF
--- a/lib/spoom/cli/bump.rb
+++ b/lib/spoom/cli/bump.rb
@@ -83,12 +83,9 @@ module Spoom
           exit(files_to_bump.empty?)
         end
 
-        output, no_errors = Sorbet.srb_tc(
-          "--no-error-sections",
-          path: exec_path,
-          capture_err: true,
-          sorbet_bin: options[:sorbet]
-        )
+        tc_args = []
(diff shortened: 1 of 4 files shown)


lib/spoom/cli/bump.rb
```diff
@@ -83,12 +83,9 @@ def bump(directory = ".")
...
         end
 
-        output, no_errors = Sorbet.srb_tc(
-          "--no-error-sections",
-          path: exec_path,
-          capture_err: true,
-          sorbet_bin: options[:sorbet]
-        )
+        tc_args = []
+        tc_args << "--no-error-sections" if Sorbet.srb_version.to_s >= "0.5.6347"
```
RyanBrushett, line 87: Bit of a nitpick but can we extract that version number to a constant?

RyanBrushett (approved):
Seems to work nicely!

vinistock:
Closing this after the discussion to keep the gemspec requirement and not change the behavior based on Sorbet's version.

PR #84, opened 2021-05-19
Allow bumping of all sigils below given at once

ghost:
I found myself wanting to bump all files currently with `typed: false` and `typed: true` to strict at once but currently this is two separate commands.

Notes:
- I'm not sure if `below` is the best keyword for this but was the one that popped in my head
- Currently this includes `typed: ignore` files, but I'm on the fence as to whether or not these should be included in this

### DIFF
--- a/README.md
+++ b/README.md
@@ -208,6 +208,18 @@ Bump the strictness from all files currently at `typed: false` to `typed: true` 
 $ spoom bump --from false --to true -f
 ```
 
+Bump the strictness from all files currently below `typed: strict` to `typed: strict` where it does not create typechecking errors:
+
+```
+$ spoom bump --to strict --all
+```
+
+Bump the strictness from all files currently below `typed: strong` and above `typed: ignore` to `typed: strong` where it does not create typechecking errors:
+
+```
+$ spoom bump --to strong --below strong --above ignore
+```
+
 Bump the strictness from a list of files (one file by line):
 
 ```
--- a/lib/spoom/sorbet/sigils.rb
+++ b/lib/spoom/sorbet/sigils.rb
@@ -84,13 +84,50 @@ module Spoom
         params(
           directory: T.any(String, Pathname),
           strictness: String,
-          extension: String
+          extension: String,
+          desired: String,
+          all: T::Boolean,
+          below: T.nilable(String),
+          above: T.nilable(String)
         ).returns(T::Array[String])
       end
-      def self.files_with_sigil_strictness(directory, strictness, extension: ".rb")
+      def self.files_with_sigil_strictness(directory, strictness, extension: ".rb",
+        desired: "true", all: false, below: nil, above: nil)
         paths = Dir.glob("#{File.expand_path(directory)}/**/*#{extension}").sort.uniq
-        paths.filter do |path|
-          file_strictness(path) == strictness
+
+        if all || below || above
+          strictness_range = sigils_range(desired: desired, below: below, above: above, all: all)
+
+          paths.filter do |path|
+            strictness_range.any?(file_strictness(path))
+          end
+        else
+          paths.filter do |path|
+            file_strictness(path) == strictness
+          end
+        end
+      end
+
+      sig do
+        params(
+          desired: String,
+          below: T.nilable(String),
+          above: T.nilable(String),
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -208,6 +208,18 @@ Bump the strictness from all files currently at `typed: false` to `typed: true`
 $ spoom bump --from false --to true -f
 ```
 
+Bump the strictness from all files except `typed: ignore` currently below `typed: strict` to `typed: strict` where it does not create typechecking errors:
```
Morriar, line 211: `--all` should really mean _all_, if you don't want _all_, use `--above`:

```suggestion
Bump the strictness from all files below `typed: strict` to `typed: strict` where it does not create typechecking errors:
```

lib/spoom/sorbet/sigils.rb
```diff
@@ -84,13 +84,49 @@ def self.change_sigil_in_files(path_list, new_strictness)
...
+          extension: String,
+          desired: String,
+          all: T::Boolean,
+          below: T.nilable(String),
+          above: T.nilable(String)
         ).returns(T::Array[String])
       end
-      def self.files_with_sigil_strictness(directory, strictness, extension: ".rb")
+      def self.files_with_sigil_strictness(directory, strictness, extension: ".rb",
+        desired: "true", all: false, below: nil, above: nil)
```
Morriar, line 95: I think it would be easier if this method was always expecting an array of strictnesses to select the files matching.

Something like:
```rb
      # finds all files in the specified directory matching `strictnesses`
      sig { params(directory: T.any(String, Pathname), strictness: T::Array[String], extension: String).returns(T::Array[String]) }
      def self.files_with_matching_strictness(directory, strictnesses, extension: ".rb")
        paths = Dir.glob("#{File.expand_path(directory)}/**/*#{extension}").sort.uniq
        paths.filter do |path|
          strictnesses.include?(file_strictness(path))
        end
      end
```

lib/spoom/sorbet/sigils.rb
```diff
@@ -84,13 +84,49 @@ def self.change_sigil_in_files(path_list, new_strictness)
...
+        params(
+          desired: String,
+          below: T.nilable(String),
+          above: T.nilable(String),
+          all: T::Boolean
+        ).returns(T::Array[String])
+      end
+      def self.sigils_range(desired:, below: nil, above: nil, all: false)
+        if all
+          VALID_STRICTNESS.take_while { |value| value != desired } - [STRICTNESS_IGNORE]
```
Morriar, line 121: `all` should really mean _all_:

```suggestion
          VALID_STRICTNESS
```

lib/spoom/sorbet/sigils.rb
```diff
@@ -84,13 +84,49 @@ def self.change_sigil_in_files(path_list, new_strictness)
...
+        elsif below && above
+          above = VALID_STRICTNESS.reverse.take_while { |value| value != above }
+          below = VALID_STRICTNESS.take_while { |value| value != below }
+          (above & below) - [desired, STRICTNESS_IGNORE]
```
Morriar, line 125: Since `VALID_STRICTNESS` is already sorted we could use indexes:

```rb
  above_index = VALID_STRICNESS.index(above)
  below_index = VALID_STRICNESS.index(below)
  range = VALID_STRICNESS[above_index..below_index]
```

lib/spoom/sorbet/sigils.rb
oneandonlymike: I don't think I agree with this one, because then that would require figuring out the range of sigils that above/below/all translate into in the `bump.rb` in order to pass them to the method. I feel like it makes more sense for that logic to live and be handled here so that `bump.rb` is just passing the options given in the input, and `sigils.rb` is translating that into what is actually needed.

Morriar: I'm not sure I understand since you can still use the method `sigil_range` from below:

```rb
strictnesses = Sigils.sigils_range(desired, below: "strong", above: "ignore", all: false)
files_to_bump = Sigils. files_with_matching_strictness(dir, strictnesses)
```

Morriar:
Marking this as stale, feel free to reopen 👍

PR #88, opened 2021-06-03
Parse --no-stdlib from config

Morriar:
When computing typing coverage for a project which use `--no-stdlib` we should pass back this option. Without it, Sorbet will hang or return a bazillion errors.

This PR makes the `Config` parser recognize `--no-stdlib` in the config file and generate it when we ask for the options string.

### DIFF
--- a/lib/spoom/sorbet/config.rb
+++ b/lib/spoom/sorbet/config.rb
@@ -29,11 +29,15 @@ module Spoom
       sig { returns(T::Array[String]) }
       attr_reader :paths, :ignore, :allowed_extensions
 
+      sig { returns(T::Boolean) }
+      attr_accessor :no_stdlib
+
       sig { void }
       def initialize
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
What exactly does `--no-stdlib` do inside Sorbet? Does it not include RBIs on objects from Ruby standard library?

Morriar:
> What exactly does `--no-stdlib` do inside Sorbet? Does it not include RBIs on objects from Ruby standard library?

By passing `--no-stdlib` you tell Sorbet to not use the RBI payload for stdlib so you're expected to provide yours as RBI files. We were using this at first in Core as a way to iterate faster. Then we got our changes merged upstream and stopped using this option.

This means that to compute the coverage timeline for the first commits in Core, we need to pass this option of Sorbet tries to load the stdlib RBIs over the payload ones.

PR #92, opened 2021-06-15
checkout the original branch instead of commit when available

oneandonlymike:
This addresses the issue that spoom would checkout the original commit instead of the original branch when running the coverage command. 

This is a draft PR as I have not figured out how to simulate being on a branch in a test yet, so do not have a test for the behaviour yet.

### DIFF
--- a/lib/spoom/cli/coverage.rb
+++ b/lib/spoom/cli/coverage.rb
@@ -44,8 +44,9 @@ module Spoom
         path = exec_path
         sorbet = options[:sorbet]
 
-        sha_before = Spoom::Git.last_commit(path: path)
-        unless sha_before
+        ref_before = Spoom::Git.current_branch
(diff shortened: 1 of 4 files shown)


lib/spoom/cli/coverage.rb
```diff
@@ -44,6 +44,7 @@ def timeline
         path = exec_path
         sorbet = options[:sorbet]
 
+        branch_before = Spoom::Git.current_branch
         sha_before = Spoom::Git.last_commit(path: path)
```
Morriar, line 48: Can we merge the two?

```suggestion
        ref_before = Spoom::Git.current_branch
        ref_before = Spoom::Git.last_commit(path: path) unless ref_before
```

Then later:

```rb
Spoom::Git.checkout(ref_before, path: path)
```

Morriar (approved):
Amazing, thanks! ❤️

RyanBrushett (approved):

PR #100, opened 2021-10-06
Fix for `requires_ancestor` error

mutecipher:
Looks like the most recent release of Sorbet introduced a breaking
change where `requires_ancestor` only accepts blocks now. This applies
the auto-correct for the issue as per https://github.com/sorbet/sorbet/pull/4691

### DIFF
--- a/lib/spoom/git.rb
+++ b/lib/spoom/git.rb
@@ -20,7 +20,7 @@ module Spoom
       err = e.read.to_s
       e.close
       i.close
-      [T.cast(out, String), T.cast(err, String), T.must(T.cast(s.value, Process::Status).success?)]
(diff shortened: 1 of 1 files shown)


lib/spoom/git.rb
```diff
@@ -20,7 +20,7 @@ def self.exec(command, *arg, path: '.')
       err = e.read.to_s
       e.close
       i.close
-      [out, err, T.cast(s.value, Process::Status).success?]
+      [T.cast(out, String), T.cast(err, String), T.must(T.cast(s.value, Process::Status).success?)]
```
Morriar, line 23: Why do we need to cast `out` and `err` here?

PR #107, opened 2021-11-03
Add Tapioca to dev dependencies

Morriar:
So we can generate RBIs for gems and keep up to date easily.

Since Tapioca is a dependency of Spoom we can't add it to the Gemspec or we'll create a dependency loop.


KaanOzkan (approved):
Was tapioca removed since adding to https://github.com/Shopify/spoom/tree/master/sorbet/rbi/gems?

vinistock (approved):

Morriar:
> Was tapioca removed since adding to https://github.com/Shopify/spoom/tree/master/sorbet/rbi/gems?

I don't think `tapioca` was ever a dependency of `spoom`. Until now I was using Tapioca from the system gems after a `gem install tapioca`.

Issue #117, opened 2022-01-31
Investigate why `test_run_srb_from_bundler_not_found` randomly fails


Morriar:
Example: https://github.com/Shopify/spoom/runs/4982433490?check_suite_focus=true

KaanOzkan:
Nit: the `assert_equal` calls in [that file](https://github.com/Shopify/spoom/blob/96634e266e2674233f3949e85ea89d4218a33da8/test/spoom/sorbet/run_test.rb#L34) has the wrong argument ordering, expected value should come first: https://docs.seattlerb.org/minitest/Minitest/Assertions.html. It causes a strange failure message in its current state.

PR #121, opened 2022-03-09
Fix position of expected values in `assert_equal`

Morriar:
As [noted](https://github.com/Shopify/spoom/issues/117#issuecomment-1025916598) by @KaanOzkan, the expected value should come first.


vinistock (approved):

RyanBrushett (approved):
<3

Issue #118, opened 2022-02-28
Spoom bump sometimes deletes empty lines


atanbt:
**Describe the bug**
When I run spoom bump sometimes it will delete the line between the sigil and the first line of the class. For example, of 28 files that it changed when I just ran it, 4 of them the sigil was unchanged but the line was deleted.

**To Reproduce**
Spoom version: 1.1.8

Run spoom bump

**Expected behavior**
It should only change files when there is a sigil change.

Morriar:
👋 Hey Alex,

Do you have a repo I can use to reproduce this problem?

Thanks

atanbt:
Hi @Morriar, unfortunately I don't. I've been seeing the issue in a private repo.

Morriar:
Is it always the same 4 files that get mangled? If yes, any change you can share their contents with me?

PR #124, opened 2022-03-31
Preserve empty lines when using `spoom bump`

bravehager:
Fix for #118

Right now, Spoom will remove empty lines after bumping files. My guess is most codebases wouldn't be affected because of enforced magic comments such as `# frozen_string_literal: true`.

Before:
```ruby
# typed: false

class Foo; end
```

After:
```ruby
# typed: true
class Foo; end
```

Adjusting the type sigil regex to only match on horizontal whitespace (tabs and spaces) seems to fix the issue, although there could be additional edge cases to consider.

### DIFF
--- a/lib/spoom/sorbet/sigils.rb
+++ b/lib/spoom/sorbet/sigils.rb
@@ -25,7 +25,7 @@ module Spoom
         STRICTNESS_INTERNAL,
       ].freeze, T::Array[String])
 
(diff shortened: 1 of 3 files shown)


Morriar:
@bravehager I just released [spoom 1.1.10](https://rubygems.org/gems/spoom/versions/1.1.10) with your fix 👍 

Thanks you again 🙏

PR #125, opened 2022-04-04
Fix encoding issues when writing non-ASCII content

Morriar:
This fixes the bug reported here: https://github.com/Shopify/tapioca/issues/881.

I traced the problem to a combination of writing the content without specifying an `encoding` [here](https://github.com/Shopify/spoom/blob/main/lib/spoom/sorbet/sigils.rb#L69) while `Encoding.default_internal` is set to `Encoding::UTF_8` by Rails [here](https://github.com/rails/rails/blob/main/railties/lib/rails.rb#L22).

In these circumstances, writing the new RBI content will fail if it contains UTF-8 characters.

### DIFF
--- a/test/spoom/sorbet/sigils_test.rb
+++ b/test/spoom/sorbet/sigils_test.rb
@@ -220,6 +220,29 @@ module Spoom
         project.destroy
       end
 
+      def test_change_sigil_in_file_with_default_internal_encoding
+        project = spoom_project
+
+        string = <<~RB
+          # typed: true
+
+          puts "À coûté 10€"
+        RB
+
+        old_encoding = Encoding.default_internal
+        project.write("file.rb", string.encode("UTF-8"))
+
+        begin
+          Encoding.default_internal = Encoding::UTF_8
+
+          Sigils.change_sigil_in_file("#{project.path}/file.rb", "strict")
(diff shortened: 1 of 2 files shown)


Morriar:
@ryanquanz this should fix your issue 🙏

test/spoom/sorbet/sigils_test.rb
```diff
@@ -220,6 +220,26 @@ def test_change_sigil_in_file_with_iso_content
...
+
+        old_encoding = Encoding.default_internal
+        Encoding.default_internal = Encoding::UTF_8
+
+        project.write("file.rb", string.encode("UTF-8"))
+        Sigils.change_sigil_in_file("#{project.path}/file.rb", "strict")
+        assert_equal("strict", Sigils.file_strictness("#{project.path}/file.rb"))
+        project.destroy
+
+        Encoding.default_internal = old_encoding
```
paracycle, line 240: Can we make this safer by putting it in an `ensure` block?
```suggestion
      ensure
        Encoding.default_internal = old_encoding if old_encoding
```
Otherwise, when/if the assertions above raise, the default internal encoding will not be restored, which might end up affecting other tests.

vinistock (approved):

PR #128, opened 2022-04-12
Do not go through all directory files when bumping

### LINKED ISSUE #127: Very slow because of file scanning, and does not respect sorbet config
**Describe the bug**
Running `spoom bump` takes a very long time for us (>2 mins on a small codebase). Root cause is we use ActiveStorage, which in dev mode creates a lot of directories and sub-directories. Spoom expands those; this line takes 62s to run. https://github.com/Shopify/spoom/blob/e1d339fa287ee5796d252cda3e7c0eb7adc6bfa7/lib/spoom/sorbet/sigils.rb#L91

Ideally I'd just ignore the storage directory, as I've done for Sorbet, but Spoom doesn't have a way to ignore those directories, [despite what the docs claim](https://github.com/Shopify/spoom#sorbet-configuration-commands).

`bundle exec spoom bump --from ignore --to false  3.44s user 34.14s system 26% cpu 2:21.30 total`

**To Reproduce**
Spoom version: master

Create lots of dirs and subdirs in tmp and run `spoom bump`

**Expected behavior**
Does not recursively scan through directories ignored in sorbet config.

Morriar:
Closes #127.

Before this patch, we were listing, reading and checking _all_ the files (direct and in subdirs) we could find from the bump `directory`. This may be highly inefficient, especially when the directory contains a lot of files that will end up ignored from the Sorbet config.

After this patch, while we still list them all, we at least do not read and check them if they get ignored by the config.

### DIFF
--- a/lib/spoom/cli/bump.rb
+++ b/lib/spoom/cli/bump.rb
@@ -58,10 +58,10 @@ module Spoom
         say("Checking files...")
 
         directory = File.expand_path(directory)
-        files_to_bump = Sorbet::Sigils.files_with_sigil_strictness(directory, from)
 
-        files_from_config = config_files(path: exec_path)
-        files_to_bump.select! { |file| files_from_config.include?(file) }
+        files_to_bump = config_files(path: exec_path)
+        files_to_bump.select! { |file|  file.start_with?(directory) && !file.end_with?(".rbi") }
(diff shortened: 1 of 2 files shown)


Morriar:
No, as pointed in the issue it only moved the problem. I'll to try a smarter way.

Morriar:
Superseded by https://github.com/Shopify/spoom/pull/322.

PR #138, opened 2022-06-14
Override user error-url-base when running Sorbet with `spoom tc`

Morriar:
Fix issues where `spoom tc` is reporting no error while `srb tc` does report errors when using `--error-url-base`.

We always override the option with a known string when running `spoom tc` so the error parser knows what to look for in the error strings.


davidstosik (approved):
Thank you! ❤️

PR #176, opened 2022-07-21
Do not swallow Sorbet errors

Morriar:
Currently, `spoom` will swallow any error from Sorbet even if it's not a type checking error.

For example, running this:

```
spoom tc --sorbet-options="--not-found"
```

Will fail silently.

To solve the problem we first [introduced a new error code in Sorbet](https://github.com/sorbet/sorbet/pull/6051) so it returns `100` if type checking finished correctly but with type checking errors.

The second step is to make Spoom check this error code so we can relay any non-type checking error to the user.

This also means requiring at least Sorbet 0.5.10187.


vinistock (approved):
Very nice

st0012 (approved):

rafaelfranca (approved):

Issue #200, opened 2022-08-29
Default strictness is not detected


mclark:
**Describe the bug**
Ruby source files which lack any sigil but are parsed by Sorbet are considered `typed: false` by default when running type checking. However, Spoom ignores them completely. This means these files can't be easily `bump`ed and they aren't represented correctly in graphs.

**To Reproduce**
Spoom version: 1.1.12

Run `spoom bump --from=false --to=true` on a directory with some ruby files lacking sigils.

**Expected behavior**
The files without sigils or Sorbet errors have `# typed: true` added to them.

---

I'd be happy to write a PR for this if it is of interest to the maintainers.

PR #201, opened 2022-08-29
support bumping files without typed sigil

mclark:
This is a proposed fix for https://github.com/Shopify/spoom/issues/200

When Sorbet encounters a file without a `typed` sigil, it treats the file the same as `typed: false`. This is also [documented](https://sorbet.org/docs/static#file-level-granularity-strictness-levels) on their website. This pull request changes Spoom to treat files with a missing sigil as `typed: false`. This was done to allow us to leverage `spoom bump` on our default typed codebase.

To accomplish this goal, I modified `Spoom::Sorbet::Sigils.file_strictness` to return the `DEFAULT_STRICTNESS` instead of `nil` when no sigil can be found in the file. I thought this might be more accurate. I also updated `Spoom::Sorbet::Sigils.change_sigil_in_files` to return a hash of paths to their previous strictnesses so that `undo` has enough information to be able to revert changes as needed. It now also takes a hash of paths to desired strictness to be consistent.

This change is pretty fundamental and will likely also affect graph output, so I'd be pretty careful here. Also, I personally have about a hot 5 minutes of experience with Spoom, so beware of rookie mistakes!

---

If reviewers think there is too much going on in this PR, it can easily be broken down into a PR that just changes the signature for `Spoom::Sorbet::Sigils.change_sigil_in_files` first and then the support for default strictness afterwards. Up to the readers, I'm happy to accommodate 🙇

### DIFF
--- a/lib/spoom/cli/bump.rb
+++ b/lib/spoom/cli/bump.rb
@@ -75,12 +75,14 @@ module Spoom
           exit(0)
         end
 
-        Sorbet::Sigils.change_sigil_in_files(files_to_bump, to)
+        update_to = files_to_bump.each_with_object({}) { |path, update_to| update_to[path] = to }
+
+        updated_from = Sorbet::Sigils.change_sigil_in_files(update_to)
 
         if force
-          print_changes(files_to_bump, command: cmd, from: from, to: to, dry: dry, path: exec_path)
-          undo_changes(files_to_bump, from) if dry
-          exit(files_to_bump.empty?)
+          print_changes(updated_from.keys, command: cmd, from: from, to: to, dry: dry, path: exec_path)
+          undo_changes(updated_from) if dry
+          exit(update_to.empty?)
         end
 
         error_url_base = Spoom::Sorbet::Errors::DEFAULT_ERROR_URL_BASE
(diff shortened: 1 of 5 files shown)


lib/spoom/cli/bump.rb
```diff
@@ -75,12 +75,14 @@ def bump(directory = ".")
           exit(0)
         end
 
-        Sorbet::Sigils.change_sigil_in_files(files_to_bump, to)
+        update_to = files_to_bump.each_with_object({}) { |path, update_to| update_to[path] = to }
```
rafaelfranca, line 78: Isn't `to` always the same value here? Why do we need tho build this hash with it then?

lib/spoom/sorbet/sigils.rb
```diff
@@ -54,32 +56,56 @@ def update_sigil(content, new_strictness)
           content.sub(SIGIL_REGEXP, sigil_string(new_strictness))
         end
 
+        sig { params(content: String, new_strictness: String).returns(String) }
```
rafaelfranca, line 59: Can you document the new methods you added? We try to document all methods in this gem.

mclark:
> We never needed this because sigils are required in our codebase using `rubocop-sorbet`

🤔 That could potentially work for us as well now.

Originally we didn't want a big explosion of `typed: false` through the whole code base when we were still in an experimental stage, but now that we are bumping to `typed: true` in large numbers, the explosion of files changed has become inevitable.

If Shopify's Sorbet team thinks this change is worthwhile I am happy to get it over the line, but I may end up experimenting with the cop's autocorrect as well. Thanks for the pointer!

PR #227, opened 2022-10-23
cli/bump: Fix "0 or 1 file(s) changed" pluralization

issyl0:
- The previous wording of "No file to bump" sounded weird given that zero is usually referred to in the plural.
- Similarly, Spoom detecting a single file to bump but referring to that single file plurally was confusing as well.

### DIFF
--- a/lib/spoom/cli/bump.rb
+++ b/lib/spoom/cli/bump.rb
@@ -132,13 +132,14 @@ module Spoom
 
       no_commands do
         def print_changes(files, command:, from: "false", to: "true", dry: false, path: File.expand_path("."))
-          if files.empty?
+          files_count = files.size
+          if files_count.zero?
             say("No files to bump from `#{from}` to `#{to}`")
             return
           end
           message = StringIO.new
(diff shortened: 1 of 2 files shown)


lib/spoom/cli/bump.rb
```diff
@@ -146,7 +146,7 @@ def print_changes(files, command:, from: "false", to: "true", dry: false, path:
             say(" + #{file_path}")
           end
           if dry && command
-            say("\nRun `#{command}` to bump them")
+            say("\nRun `#{command}` to bump #{files.length > 1 ? "them" : "it"}")
```
Morriar, line 149: Thanks.

Do you mind changing [line 141](https://github.com/Shopify/spoom/blob/9a687f707a85336a86ce6bc930d3dd1421dff642/lib/spoom/cli/bump.rb#L141) to also use `length` so we keep some uniformity please?

issyl0: Done! Thanks for the review!

PR #230, opened 2022-10-27
Collect metrics on non rbi files

KaanOzkan:
Allows for more finer grained processing and a better indication of typing work done on a project.

### DIFF
--- a/lib/spoom/coverage.rb
+++ b/lib/spoom/coverage.rb
@@ -19,13 +19,24 @@ module Spoom
 
         new_config = config.copy
         new_config.allowed_extensions.reject! { |ext| !rbi && ext == ".rbi" }
-
-        metrics = Spoom::Sorbet.srb_metrics(
+        flags = [
           "--no-config",
           "--no-error-sections",
           "--no-error-count",
           "--isolate-error-code=0",
           new_config.options_string,
+        ]
+
+        metrics = Spoom::Sorbet.srb_metrics(
+          *flags,
+          path: path,
+          capture_err: true,
+          sorbet_bin: sorbet_bin,
+        )
+        # Collect extra information using a different configuration
+        flags << "--ignore sorbet/rbi/"
+        metrics_without_rbis = Spoom::Sorbet.srb_metrics(
+          *flags,
           path: path,
           capture_err: true,
           sorbet_bin: sorbet_bin,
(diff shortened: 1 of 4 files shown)


vinistock (approved):

test/spoom/snapshot_test.rb
```diff
@@ -72,6 +73,7 @@ def test_snapshot_project_without_rbi
         assert_equal(8, snapshot.methods_without_sig)
         assert_equal(6, snapshot.calls_typed)
         assert_equal(1, snapshot.calls_untyped)
+        assert_equal(1, snapshot.non_generated_methods_with_sig)
```
KaanOzkan, line 76: Added simple tests that I forgot earlier.

lib/spoom/coverage.rb
```diff
@@ -19,13 +19,24 @@ def snapshot(path: ".", rbi: true, sorbet_bin: nil)
...
+
+        metrics = Spoom::Sorbet.srb_metrics(
+          *flags,
+          path: path,
+          capture_err: true,
+          sorbet_bin: sorbet_bin,
+        )
+        # Collect extra information using a different configuration
+        flags << "--ignore sorbet/rbi/gems sorbet/rbi/dsl"
+        secondary_metrics = Spoom::Sorbet.srb_metrics(
```
Morriar, line 38: What do you think about `metrics_without_rbis` instead?

lib/spoom/coverage/snapshot.rb
```diff
@@ -22,6 +22,8 @@ class Snapshot < T::Struct
       prop :calls_untyped, Integer, default: 0
       prop :calls_typed, Integer, default: 0
       prop :sigils, T::Hash[String, Integer], default: Hash.new(0)
+      prop :non_generated_methods_with_sig, Integer, default: 0
```
Morriar, line 25: I'm not a fan of this name either. What do you think about `methods_with_sig_excluding_rbis`?

lib/spoom/coverage.rb
KaanOzkan: It's less accurate but slightly better. I was thinking this could be used further but for now `secondary` isn't informative enough.

lib/spoom/coverage/snapshot.rb
```diff
@@ -22,6 +22,8 @@ class Snapshot < T::Struct
       prop :calls_untyped, Integer, default: 0
       prop :calls_typed, Integer, default: 0
       prop :sigils, T::Hash[String, Integer], default: Hash.new(0)
+      prop :methods_with_sig_excluding_rbis, Integer, default: 0
+      prop :non_generated_sigils, T::Hash[String, Integer], default: Hash.new(0)
```
Morriar, line 26:
```suggestion
      prop :sigils_excluding_rbis, T::Hash[String, Integer], default: Hash.new(0)
```

KaanOzkan: It might confuse people because we aren't excluding all `.rbi` files but I don't have strong opinions.

Morriar (approved):

Issue #259, opened 2022-12-15
Spoom should raise if it can't produce snapshot metrics


Morriar:
If the Sorbet process gets killed while processing the metrics, Spoom will just return a bunch of metrics at 0s leaving the user clueless that something went terribly wrong.

* [`Sorbet.srb_metrics`](https://github.com/Shopify/spoom/blob/main/lib/spoom/sorbet.rb#L96-L109) doesn't check for the command status will just return `nil` if it can't read the metrics file (we should make it raise).
* [`Coverage.snapshot`](https://github.com/Shopify/spoom/blob/main/lib/spoom/coverage.rb#L45-L46) will just return an empty `Snapshot` (so all values at `0`) if the metrics object returned by `srb_metrics` is `nil` (we should make it bubble the exception).

bitwise-aiden:
Closed by above PR chain

PR #273, opened 2023-01-13
Add raise to sorbet metrics

bitwise-aiden:
## Description 

Closes: https://github.com/Shopify/spoom/issues/259

This PR updates `#srb_metrics` to raise an error based on the output captured from the call to Sorbet. 

I've also removed `capture_err` from `#srb_metrics` since the method doesn't return the exec result, so to my understanding the caller doesn't need to be opinionated about how it's called. 
This has also meant that it is possible to capture the specific error message and bubble it up to the user. 

## What should reviewers focus on?

- Do we want to raise the error message directly from Sorbet?
- Any objections to removing `capture_err` from the `#srb_metrics` method?
- Do we want to wrap the error in coverage to return something nicer to the user?
  - Currently it will bubble the error message produced by Sorbet.


lib/spoom/sorbet.rb
```diff
@@ -89,21 +93,23 @@ def srb_version(*arg, path: ".", capture_err: false, sorbet_bin: nil)
...
           "--metrics-file",
           metrics_file,
           *arg,
           path: path,
-          capture_err: capture_err,
+          capture_err: true,
           sorbet_bin: sorbet_bin,
         )
+
+        raise Error::Metrics, result.err unless result.status
```
Morriar, line 111: Doesn't that mean we're going to raise for any type error? I think we want to raise only if the exit code == 137 (killed)? 

There is already a [constant to do something similar on segfaults](https://github.com/Shopify/spoom/blob/main/lib/spoom/sorbet.rb#L18). I think we could move all this logic to the [`srb` method](https://github.com/Shopify/spoom/blob/main/lib/spoom/sorbet.rb#L31) so it raises exceptions accordingly and thus clean the mess in https://github.com/Shopify/spoom/blob/main/lib/spoom/cli/bump.rb#L112-L118.

test/spoom/cli/coverage_test.rb
```diff
@@ -489,6 +489,17 @@ def test_finish_on_original_branch
         assert_equal("fake-branch", @project.git_current_branch)
       end
 
+      def test_error_raised_when_metrics_fails
```
Morriar, line 492: Is this test still relevant?

bitwise-aiden: Nope, just cleaning that up as we speak ;p

Morriar (commented):
One comment about the structure of these changes.

I think this PR can actually be cut into 3 smaller, easier to read PRs:

1. The Gemfile + RBI changes
2. The refactor around `segfault` (and related changes to the `run` and `bump` commands)
3. The new feature for `killed`

bitwise-aiden:
I'll split things out into their own PRs tomorrow

PR #277, opened 2023-01-17
Update handling of Sorbet errors

bitwise-aiden:
## Description

As a follow on from [this PR](https://github.com/Shopify/spoom/pull/273), I've updated the call sites that were handling segfaults to now handle the exception being raised.


Morriar:
Looking at the test failures I think we also need to update the `bump` command in the same PR.

I like the refacftor though! ❤️

bitwise-aiden:
Closing as handling of errors needs to be in the previous PR

PR #279, opened 2023-01-18
Bitwise aiden/update segfault handling

bitwise-aiden:
## Description 

This is PR 1 of 3 to address the issue outlined in this issue: https://github.com/Shopify/spoom/issues/259

1. https://github.com/Shopify/spoom/pull/278
1. **Bitwise aiden/update segfault handling**
1. TODO

In this PR I've added handling in the `Spoom::Sorbet#srb` method to check for the error code being segfault, raising if so. It will then pass out the `ExecResult` so the caller can inspect things like `#err` if necessary. 

Because we now raise in `#srb` I've updated the error handling through the rest of the application to catch the error and perform the existing handling of segfaults.  

## Approach 

Talking with @Morriar and @KaanOzkan the approach that was landed on was to catch the issue at the Sorbet level and raise an explicit exception for Segfault (exit code: 139) and eventually Killed (exit code: 137). This means that the current handling where we check for segfault can be repalced with rescuing and then outputting the error message that currently exists.


Morriar (approved):

lib/spoom/sorbet.rb
```diff
@@ -34,7 +55,14 @@ def srb(*arg, path: ".", capture_err: false, sorbet_bin: nil)
         else
           arg.prepend("bundle", "exec", "srb")
         end
-        Spoom.exec(*T.unsafe(arg), path: path, capture_err: capture_err)
+        result = Spoom.exec(*T.unsafe(arg), path: path, capture_err: capture_err)
+
+        case result.exit_code
+        when SEGFAULT_CODE
+          raise Error::Segfault.new("Sorbet segfaulted.", result)
+        end
```
vinistock, line 63: NIT: do we need a case if we only have a single when? How about switching to an if statement?

vinistock (approved):

PR #303, opened 2023-02-28
Some cleaning in the Git helpers

Morriar:
Contains 3 changes:

1. Make Exec.result contain `<not captured>` if Spoom.exec command was ran without `capture_err`.
2. Make Spoom.exec use `capture_err` by default, since we return an ExecResult the sensible default is for it to contain the error output.
3. Remove useless method `Spoom::Git.exec` that is basically a duplication of `Spoom.exec`.


lib/spoom.rb
```diff
@@ -55,7 +55,7 @@ def exec(cmd, *arg, path: ".", capture_err: false)
         stdout, status = T.unsafe(Open3).capture2([cmd, *arg].join(" "), chdir: path)
         ExecResult.new(
           out: stdout,
-          err: "",
+          err: "<not captured>",
```
vinistock, line 58: What do you think about using `nil` instead? I feel like that would better inform clients that they need to do something when there's no error captured.

dirceu (approved):

PR #172, opened 2022-07-13
Add `Context` abstraction

Morriar:
Basically what `TestHelpers::Project` was but with a more generic angle.

A context represents a directory (that can be the root directory of a project).

On a context, one can:
* manipulate files
* run commands (bundle, sorbet, arbitrary command)
* deal with project specific stuffs (Gemfile, sorbet/config, etc.)
* ... the scope can be extended as needed

Other projects such as Tapioca or RBI-Central can re-use Context to implement their own logic.

I only migrated the tests to get a look and feel of the thing.
To migrate the different commands to use a context it will take a bit of refactor I'll do in another PR.

PR #307, opened 2023-02-28
Move all Sorbet helpers as methods on Context

Morriar:
Follow up on #306, move all the `Sorbet.*` helpers as methods on the `Context` under `Context::Sorbet` as to removes even more duplication introduced during #172 and provide a single way of manipulating contexts.


KaanOzkan (commented):
What were you thinking for usages such as this https://github.com/Shopify/tapioca/blob/d96541002909dc4efcb6ee9273e88e8498b13a45/lib/tapioca/helpers/sorbet_helper.rb#L32? A global `Context` instance or individual ones in each helper instance?

Morriar:
> What were you thinking, a global Context instance or individual ones in each helper instance?

Creating a context is pretty low cost: `Context.new(".")` doesn't do any FS operation, it just creates an instance so we could go either way. To avoid refactoring too much around Tapioca we could just do:

```rb
context = Context.new(".")
context.srb(sorbet_args.join(" "), sorbet_bin: sorbet_path, capture_err: true)
```

egiurleo (approved):

PR #308, opened 2023-02-28
CLI commands use Context

Morriar:
Last follow up to https://github.com/Shopify/code-db/pull/575, make the CLI commands use Context since most of what is needed to run the commands is now available on the Context.


KaanOzkan (approved):

egiurleo (approved):
Nice! This is so clean 😍

PR #353, opened 2023-05-09
Remove Ruby LSP from Gemfile

andyw8:
Was originally added in https://github.com/Shopify/spoom/pull/150 but now Ruby LSP can be used without it being in the Gemfile.

(and removing it means one less things for Dependabot to notify about).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gemspec
 
 group :development do
   gem "debug"
-  gem "ruby-lsp"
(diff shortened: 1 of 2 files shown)


KaanOzkan (approved):
VSCode users can get it by using https://github.com/Shopify/vscode-ruby-lsp.

For other editors you'd have to insert `ruby-lsp` back in the bundle, `gem install ruby-lsp`

PR #362, opened 2023-05-23
Fix broken CLI test output

egiurleo:
The output for one of the CLI tests has some incorrect spacing, which is causing CI to fail.

### DIFF
--- a/test/spoom/cli/cli_test.rb
+++ b/test/spoom/cli/cli_test.rb
@@ -34,10 +34,10 @@ module Spoom
             spoom tc              # Run Sorbet and parses its output
 
           Options:
-                [--color], [--no-color]  # Use colors
-                                         # Default: true
-            p, [--path=PATH]             # Run spoom in a specific path
-                                         # Default: .
+               [--color], [--no-color]  # Use colors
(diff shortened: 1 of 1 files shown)


Morriar:
Could it be related to the Thor bump made in https://github.com/Shopify/spoom/pull/355?

egiurleo:
@Morriar I don't think so, because even though the PR with the thor bump was opened a while ago, it wasn't merged until this morning, after I fixed the test! I've also tried changing thor versions on a commit with the failing test, and the test continues to fail on Spin with thor 1.2.1 and 1.2.2.

I ran a `git bisect` on Spin to see if I could find the origin of the failing test and landed on one of the commits in [this PR](https://github.com/Shopify/spoom/pull/172/commits) from August 2022 -- the test reliably fails both on Spin and on my machine from that commit all the way up to this morning. However, that makes no sense to me, because it was never caught on CI? So either the test was/is flaky, or CI was somehow changing something about the environment that was causing it to pass.

I don't know if it's worth investigating further, but I can keep an eye on Spoom CI and see if any other flaky behavior comes up.

PR #374, opened 2023-05-31
Remove Ruby 2.7 support

Morriar:
[Ruby 2.7 is now EOL](https://www.ruby-lang.org/en/downloads/branches/#:~:text=03%2D31%20(expected)-,Ruby%202.7,-status%3A%20eol%0Arelease).

This PR removes support for it.


vinistock (commented):
My take would be waiting a little longer. There might be a tail of applications still on 2.7 and the effort to maintain support for it is minimal

PR #375, opened 2023-05-31
Deadcode: handle ERB files

Morriar:
Follow up on #373, this PR adds support for ERB files so we can detect dead code in Rails apps.

Rails actually uses a superset of ERB to implement the templates. This is handled thanks to a [custom Erubi engine](https://github.com/rails/rails/blob/main/actionview/lib/action_view/template/handlers/erb/erubi.rb). To support the same ERB than Rails without actually depending on Rails we just copy the Erubi engine inside Spoom.


bitwise-aiden (approved):

lib/spoom/deadcode/erb.rb
```diff
@@ -0,0 +1,82 @@
+# typed: true
+# frozen_string_literal: true
+
+module Spoom
+  module Deadcode
+    # Custom engine to handle ERB templates as used by Rails.
+    # Copied from https://github.com/rails/rails/blob/main/actionview/lib/action_view/template/handlers/erb/erubi.rb.
```
rafaelfranca, line 7: Did you copy the entire file? If so you need to copy the license as well.

Issue #379, opened 2023-06-02
Spoom bump not ignoring files as expected


plvaldes:
**Describe the bug**
The command `spoom bump --from false --to true --dry` is trying to bump files that I have in my ignore list in `sorbet/config`.

I am assuming that spoom config is picked from `sorbet/config` as I read in the docs. Actually, running `bin/spoom config` in my project lists all the ignore paths in that file (see link in steps to reproduce)

**To Reproduce**
Spoom version: '1.2.1'

There are more details about the context and what I did [in this discourse thread](https://discourse.shopify.io/t/solid-track-same-pr-generated-after-shipping/38336/7?u=pablo.valdes), but the TL;DR should be:
- in a file that has `# typed: true`, change it `false`
- add that file to `sorbet/config` as `--ignore=the_file.rb`
- run `spoom bump --from false --to true --dry`

**Expected behavior**
The file should be ignored and no errors should be displayed

Morriar:
👋 Hey @plvaldes, 

Do you mind posting your Sorbet config here please? That would help me fix the issue 🙏

plvaldes:
Hey @Morriar ! 👋🏼 This is the file:

```
--dir
.
--ignore=/config/routes.rb
--ignore=/app/controllers/api
--ignore=/lib/generators/
--ignore=/test/lib/generators
```

It's in the [melody repo](https://github.com/Shopify/melody/blob/main/sorbet/config)

PR #383, opened 2023-06-06
Do not list files ignored by Sorbet

Morriar:
We ignore files matching the paths ignored in the `sorbet/config` and try to reproduce Sorbet behvior.

From Sorbet docs on `--ignore`:
> Ignores input files that contain the given string in their paths (relative to the input path passed to
> Sorbet). Strings beginning with / match against the prefix of these relative paths; others are substring
> matchs. Matches must be against whole folder and file names, so `foo` matches `/foo/bar.rb` and
> `/bar/foo/baz.rb` but not `/foo.rb` or `/foo2/bar.rb`.

To produce the same behavior, we match files using `File.fnmatch?` and we were prefixing and suffixing the paths with `**`.

When the Sorbet config contains ignored files with extensions like this:

```
--ignore=foo.rb
```

This doesn't work as we are trying to match `foo.rb` with `**/foo.rb/**`.

To fix this problem, we only append `**` to the ignored path IF it does contain a file extension.

Fixes https://github.com/Shopify/spoom/issues/379.

### DIFF
--- a/lib/spoom/context/sorbet.rb
+++ b/lib/spoom/context/sorbet.rb
@@ -69,7 +69,24 @@ module Spoom
         allowed_extensions = Spoom::Sorbet::Config::DEFAULT_ALLOWED_EXTENSIONS if allowed_extensions.empty?
         allowed_extensions -= [".rbi"] unless include_rbis
 
-        excluded_patterns = config.ignore.map { |string| File.join("**", string, "**") }
+        excluded_patterns = config.ignore.map do |string|
+          # We need to simulate the behavior of Sorbet's `--ignore` flag.
+          #
+          # From Sorbet docs on `--ignore`:
+          # > Ignores input files that contain the given string in their paths (relative to the input path passed to
+          # > Sorbet). Strings beginning with / match against the prefix of these relative paths; others are substring
+          # > matchs. Matches must be against whole folder and file names, so `foo` matches `/foo/bar.rb` and
+          # > `/bar/foo/baz.rb` but not `/foo.rb` or `/foo2/bar.rb`.
+          string = if string.start_with?("/")
+            # Strings beginning with / match against the prefix of these relative paths
+            File.join(absolute_path, string)
+          else
+            # Others are substring matchs
+            File.join(absolute_path, "**", string)
+          end
+          # Matches must be against whole folder and file names
+          "#{string.delete_suffix("/")}{,/**}"
+        end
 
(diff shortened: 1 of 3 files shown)


test/spoom/context/sorbet_test.rb
```diff
@@ -339,6 +341,25 @@ def test_context_srb_files_with_custom_ignore
...
+        context.write_sorbet_config!(<<~CONFIG)
+          .
+          --ignore=/config/routes.rb
+          --ignore=/lib/generators/
+        CONFIG
+
+        context.write!("foo.rb", "")
+        context.write!("config/routes.rb", "")
+        context.write!("lib/generators/generator1.rb", "")
+        context.write!("lib/generators/generator2.rb", "")
```
paracycle, line 356: Per my comment above, what happens if we add a file named `/foo/lib/generators/bar.rb` here?

Morriar: I added this test case 👍

lib/spoom/context/sorbet.rb
```diff
@@ -69,7 +69,24 @@ def srb_files(with_config: nil, include_rbis: true)
...
+          # > Ignores input files that contain the given string in their paths (relative to the input path passed to
+          # > Sorbet). Strings beginning with / match against the prefix of these relative paths; others are substring
+          # > matchs. Matches must be against whole folder and file names, so `foo` matches `/foo/bar.rb` and
+          # > `/bar/foo/baz.rb` but not `/foo.rb` or `/foo2/bar.rb`.
+          string = if string.start_with?("/")
+            # Strings beginning with / match against the prefix of these relative paths
+            File.join(absolute_path, string)
+          else
+            # Others are substring matchs
+            File.join("**", string)
```
paracycle, line 85: I wonder if we should add the `absolute_path` here as well. Otherwise, could we not risk matching `string` against a part of the absolute path too? Like, if the absolute path of our project is `/foo/bar/baz` and the ignore was `bar`, `**/bar/**` would match all the files, right?

Morriar: I added it but I'm started to wonder if the `FileCollector` should work with relative paths instead.

At the moment we're making all the paths absolutes, then we visit with `FileCollector`, then we strip the absolute path root from the results. This back and forth seems unnecessary.

Let's see. If the tests pass I can merge this then revisit later.

paracycle (approved):

lib/spoom/context/sorbet.rb
paracycle: Agreed, doing relative matches would definitely be nicer.

PR #420, opened 2023-07-19
Introduce dead code plugin base

Morriar:
This PR introduces the base concept for dead code plugins.

The idea is pretty simple, as the indexer visits the code it adds the definitions it finds to the index then calls the plugins to let them ignore those definitions or add references to them.

Here's a very simple example of plugin to ignore a class that has been indexed:

```rb
class MyPlugin < Spoom::Deadcode::Plugins::Base
  def on_define_class(indexer, definition)
    definition.ignored! if definition.name == "Foo"
  end
end
```

This plugin will be called then the indexer processes the `class Foo; end` node. It will call the `on_define_class` from each registered plugin and pass it the indexed `Foo` definition. Note that when `on_define_class` the definition has already been indexed.

Plugins can also be used to handle DSLs and meta-programming.
Let's consider this DSL for example:

```rb
class Foo
  will_call :foo

  def foo; end
end
```

Through meta-programming, the `Foo` class will call the `foo` method so we want to mark it as alive.
Here's how it can be done in the plugin:

```rb
class MyPlugin < Spoom::Deadcode::Plugins::Base
  def on_send(indexer, send)
    return unless send.name == "will_call"
    return if send.args.empty?
    
    method_name = indexer.node_string(send.args.first).delete_prefix(":")
    indexer.reference_method(method_name, send.node)
  end
end
```

### DIFF
--- a/lib/spoom/deadcode/plugins/base.rb
+++ b/lib/spoom/deadcode/plugins/base.rb
@@ -1,0 +1,201 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "set"
+
+module Spoom
+  module Deadcode
+    module Plugins
+      class Base
+        extend T::Sig
+        extend T::Helpers
+
+        abstract!
+
+        class << self
+          extend T::Sig
+
+          # Plugins DSL
+
+          # Mark methods matching `names` as ignored.
+          #
+          # Names can be either strings or regexps:
+          #
+          # ~~~rb
+          # class MyPlugin < Spoom::Deadcode::Plugins::Base
+          #   ignore_method_names(
+          #     "foo",
+          #     "bar",
+          #     /baz.*/,
+          #   )
+          # end
+          # ~~~
+          sig { params(names: T.any(String, Regexp)).void }
+          def ignore_method_names(*names)
+            save_names_and_patterns(names, :@ignored_method_names, :@ignored_method_patterns)
+          end
+
+          private
+
+          sig { params(names: T::Array[T.any(String, Regexp)], names_variable: Symbol, patterns_variable: Symbol).void }
+          def save_names_and_patterns(names, names_variable, patterns_variable)
+            ignored_names = instance_variable_set(names_variable, Set.new)
+            ignored_patterns = instance_variable_set(patterns_variable, [])
+
+            names.each do |name|
+              case name
+              when String
+                ignored_names << name
+              when Regexp
+                ignored_patterns << name
+              end
+            end
+          end
+        end
+
+        # Indexing event methods
+
+        # Called when an accessor is defined.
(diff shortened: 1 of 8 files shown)


lib/spoom/deadcode/plugins/base.rb
```diff
@@ -0,0 +1,201 @@
...
+          sig { params(names: T.any(String, Regexp)).void }
+          def ignore_method_names(*names)
+            save_names_and_patterns(names, :@ignored_method_names, :@ignored_method_patterns)
+          end
+
+          private
+
+          sig { params(names: T::Array[T.any(String, Regexp)], names_variable: Symbol, patterns_variable: Symbol).void }
+          def save_names_and_patterns(names, names_variable, patterns_variable)
+            ignored_names = instance_variable_set(names_variable, Set.new)
```
vinistock, line 42: Why do we need `instance_variable_set` instead of just referencing the instance variable? Is it because Sorbet complains that it's not declared?

If that's the case, would it be worth using `typed: true` just for this file for better readability?

vinistock (approved):
In many aspects, this is very similar to the extensions system in the Ruby LSP. Some ideas that were useful to us and could be useful here too:

1. Is it worth treating even the base functionality in the same way a plugin is treated? Would that clean up the code?
2. Instead of defining empty methods that are always invoked, could we register plugins and only invoke the ones that registered for a specific event? That may not seem like much, but in a deeply nested AST it saves quite a few method dispatches and may improve performance

Finally, I know this is unrelated to this PR and I appreciate the decision to put the dead code analysis in Spoom to reduce maintenance cost of yet another gem.

However, this seems incredibly useful and completely independent from the other Spoom responsibilities, which are mostly related to Sorbet. I wonder if in the future we can extract this to a separate gem.

Morriar:
> Is it worth treating even the base functionality in the same way a plugin is treated? Would that clean up the code?

Do you mean what the indexer is doing? It a slightly different behavior, the indexer is responsible of visiting the tree and trigger the `on_*` events from the plugins.

> Instead of defining empty methods that are always invoked, could we register plugins and only invoke the ones that registered for a specific event? That may not seem like much, but in a deeply nested AST it saves quite a few method dispatches and may improve performance

Yeah, it seems like a good improvement but in practice when running this on core with 30 plugins vs. 0 plugins it doesn't impact much the analysis time. As we get more plugins this will become more significant.

> However, this seems incredibly useful and completely independent from the other Spoom responsibilities, which are mostly related to Sorbet. I wonder if in the future we can extract this to a separate gem.

Yeah we discussed this in the past, I'm not sure we're willing to maintain yet another gem at the moment. Maybe we can change this closer from the "official release".

PR #437, opened 2023-08-14
Ignore constants, classes and modules from dead code plugins

Morriar:
On top of #420, this PR introduces 3 new DSL methods for dead code plugins:

* `ignore_class_names`
* `ignore_constant_names`
* `ignore_module_names`


egiurleo (approved):
Nice!

PR #440, opened 2023-08-14
Introduce conditional ignore DSL methods for dead code plugins

Morriar:
This PR introduces a bunch of DSL methods to ignore candidates:

* `ignore_constants_if`
* `ignore_classes_if`
* `ignore_methods_if`
* `ignore_modules_if`

They can be used in dead code plugin to conditionally ignore candidates:

```rb
module Spoom
  module Deadcode
    module Plugins
      class MyPlugin < Base
        ignore_classes_if { |indexer, _definition| indexer.path.match?(/foo\.rb/) }
      end
    end
  end
end
```


egiurleo (approved):
This is so cool! I imagine this will also help us improve the `ignore_xxx` API as we see which custom blocks get used and identify patterns among them.

PR #441, opened 2023-08-15
Indexer calls internal methods of plugin base

Morriar:
By making the indexer call the `internal_on_*` method rather than `on_*` directly, we make it easier for plugin implementors to override the public API without breaking internal behavior.

Without this change, if a plugin overrides `on_define_class` and forgets to call `super`, it would break the `ignore_class_names` DSL.

### DIFF
--- a/lib/spoom/deadcode/indexer.rb
+++ b/lib/spoom/deadcode/indexer.rb
@@ -230,7 +230,7 @@ module Spoom
           end
         else
           @plugins.each do |plugin|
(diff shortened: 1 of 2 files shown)


bitwise-aiden (approved):
Ahh, nice change!

egiurleo (approved):
Oooh yes, and then this opens the door for people to be able to add custom rules by overriding `on_define_xxx`. Nice!

st0012 (approved):

PR #442, opened 2023-08-15
Rename `ignore_class_names` into `ignore_classes_named`

Morriar:
And similar for constants, methods and modules.

As discussed in https://github.com/Shopify/spoom/pull/438#discussion_r1294799135.


egiurleo (approved):
Thank you 😍

PR #452, opened 2023-08-29
Introduce Spoom::Deadcode::Remover

Morriar:
This class provides the behavior required to delete code that is deemed dead by the `Spoom::Deadcode` process.

It does best effort to keep the right formatting and limit the diff in the edited file hence the different cases we need to support around new lines and parenthesis.


lib/spoom/deadcode/remover.rb
```diff
@@ -0,0 +1,616 @@
...
+
+            new_line_before = node_before && send_context.node.location.start_line - node_before.location.end_line < 2
+            new_line_after = node_after && node_after.location.start_line - send_context.node.location.end_line <= 2
+          else
+            at_line = send_context.node.location.end_line
+            new_line_before = true
+            new_line_after = node_after && node_after.location.start_line - send_context.node.location.end_line < 2
+          end
+
+          lines_to_insert = []
```
vinistock, line 238: NIT: since we're not using any characters to join, we could just do a string here
```suggestion
          lines_to_insert = +""
```

vinistock (approved):

Issue #456, opened 2023-09-01
Minitest backtrace filter


andyw8:
As discussed with @Morriar:

The Minitest backtrace filter added to ruby-lsp in https://github.com/Shopify/ruby-lsp/pull/959 could potentially ship with Spoom, since it would be useful for other projects.

PR #457, opened 2023-09-01
Add minitest backtrace filter

andyw8:
Closes https://github.com/Shopify/spoom/issues/456

### DIFF
--- a/README.md
+++ b/README.md
@@ -348,6 +348,16 @@ Find all the symbols for a file:
 puts client.document_symbols("file://path/to/my/file.rb")
 ```
 
+### Backtrace Filtering
+
+Spoom provides a backtrace filter for Minitest to remove the Sorbet frames from test failures, giving a more readable output. To enable it:
+
+```ruby
+# test/test_helper.rb
+require "spoom/backtrace_filter/minitest"
+Minitest.backtrace_filter = Spoom::BacktraceFilter::Minitest.new
+```
+
 ## Development
 
 After checking out the repo, run `bin/setup` to install dependencies. Then, run `bin/test` to run the tests. You can also run `bin/console` for an interactive prompt that will allow you to experiment. Don't forget to run `bin/sanity` before pushing your changes.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -348,6 +348,16 @@ Find all the symbols for a file:
 puts client.document_symbols("file://path/to/my/file.rb")
 ```
 
+### Backtrace Filtering
+
+Spoom provides a backtrace filter for Minitest to remove the Sorbet frames from test failures, giving a more readable output. To enable it:
+
+```ruby
+# test/test_helper.rb
+require "spoom/minitest/backtrace_filter"
```
andyw8, line 357: (wondering where this should be a manual require or not)

lib/spoom/minitest/backtrace_filter.rb
```diff
@@ -0,0 +1,21 @@
+# typed: strict
+# frozen_string_literal: true
+
+require "minitest"
+
+module Spoom
+  module BacktraceFilter
+    class Minitest < ::Minitest::BacktraceFilter
```
andyw8, line 8: I originally had `Spoom::Minitest::BacktraceFilter` but that causes typecheck problems with tests in the `Spoom` namespace that inherit from `Minitest::Test`.

lib/spoom.rb
```diff
@@ -18,4 +18,5 @@ class Error < StandardError; end
 require "spoom/deadcode"
 require "spoom/sorbet"
 require "spoom/cli"
+require "spoom/minitest/backtrace_filter"
```
vinistock, line 21: I'd make this require manual and not included in the default require.

lib/spoom/minitest/backtrace_filter.rb
andyw8: I think this way is ok, have renamed the file to match.

vinistock (approved):
I think this is fine, but I'll let Alex decide if he prefers the default require

Morriar (approved):

PR #463, opened 2023-09-15
Bump version to v1.2.4

andyw8:


andyw8:
(also updated the RBIs since they were a little stale)

PR #468, opened 2023-09-20
Mark methods called by MiniTest.setup as alive

Morriar:
cc. @kmcphillips


kmcphillips (approved):
Nice. Yes I think this is good. We use `ActiveSupport::TestCase` which is a subclass of `Minitest::Test`.

Morriar:
Indeed, it's a ActiveSupport::TestCase and not a Minitest one 🤦 

I moved the rule to the right file and also handled `teardown` as @andyw8 pointed out.

greatscotty:
Accidentally hit this branch. Please make sure to rebase this branch to the newest version before merging

PR #520, opened 2024-03-01
Delete some useless commands

Morriar:
The commands `spoom config` and `spoom files` are not useful, let's delete them.


Morriar:
Thanks @andyw8 I didn't know about this docs and it points to an interesting feature: how to generate proper metrics for a componentized project 🤔

PR #524, opened 2024-03-06
Move all Sorbet related commands as subcommands

Morriar:
Currently, all the commands are top level making both the code and DX a bit messy. I also want to make some space for other commands such as the `deadcode` related ones.

Those are the commands currently merged:

```
  spoom bump            # Bump Sorbet sigils from `false` to `true` when no errors
  spoom coverage        # Collect metrics related to Sorbet coverage
  spoom lsp             # Send LSP requests to Sorbet
  spoom tc              # Run Sorbet and parses its output
```

With this change, all commands related to Sorbet are moved under the parent command `srb`:

```
  spoom srb bump            # Bump Sorbet sigils from `false` to `true` when no errors
  spoom srb coverage        # Collect metrics related to Sorbet coverage
  spoom srb lsp             # Send LSP requests to Sorbet
  spoom srb tc              # Run Sorbet and parses its output
```

To avoid breaking right now, I kept the top level commands to redirect to the `srb` ones with a deprecation message:

```
$ bundle exec spoom tc
Warning: This command is deprecated. Please use spoom srb tc instead.
```

We can remove them in the later major version bump.


lib/spoom/cli.rb
```diff
@@ -21,17 +17,81 @@ class Main < Thor
...
+        type: :string,
+        desc: "Command to suggest if files can be bumped"
+      option :count_errors,
+        type: :boolean,
+        default: false,
+        desc: "Count the number of errors if all files were bumped"
+      option :sorbet_options, type: :string, default: "", desc: "Pass options to Sorbet"
+      sig { params(directory: String).void }
+      def bump(directory = ".")
+        say_warning("This command is deprecated. Please use `spoom srb bump` instead.")
```
andyw8, line 58: People may wonder thew new one is the same or not. Would be better to say it's been 'renamed' or 'moved'? Or we could append, "...instead, which has identical behavior".

Morriar: Do you think the release notes would be enough?

andyw8: Maybe, but I suspect few people read them.

PR #527, opened 2024-03-13
Migrate dead code detection to Prism

Morriar:
Replace the dependency on SyntaxTree by Prism.

This PR may be easier to read commit by commit as switching the parser deeply impacted the implementation and required a lot of changes. I used the tests as source of truth.

### DIFF
--- a/lib/spoom/deadcode.rb
+++ b/lib/spoom/deadcode.rb
@@ -43,11 +43,13 @@ module Spoom
       def parse_ruby(ruby, file:)
         result = Prism.parse(ruby)
         unless result.success?
-          message = result.errors.map do |e|
-            "#{e.message} (at #{e.location.start_line}:#{e.location.start_column})."
-          end.join(" ")
+          message = +"Error while parsing #{file}:\n"
 
-          raise ParserError, "Error while parsing #{file}: #{message}"
+          result.errors.each do |e|
+            message << "- #{e.message} (at #{e.location.start_line}:#{e.location.start_column})\n"
+          end
+
+          raise ParserError, message
         end
 
         result.value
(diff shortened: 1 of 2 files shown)


lib/spoom/deadcode.rb
```diff
@@ -30,17 +35,21 @@ def initialize(message, parent:)
...
       sig { params(ruby: String, file: String).returns(SyntaxTree::Node) }
       def parse_ruby(ruby, file:)
-        SyntaxTree.parse(ruby)
-      rescue SyntaxTree::Parser::ParseError => e
-        raise ParserError.new("Error while parsing #{file} (#{e.message} at #{e.lineno}:#{e.column})", parent: e)
+        result = Prism.parse(ruby)
+        unless result.success?
+          message = result.errors.map do |e|
+            "#{e.message} (at #{e.location.start_line}:#{e.location.start_column})."
+          end.join(" ")
```
paracycle, line 48: It seems like you are trying to display multiple errors in a single line here, and I think the resulting text ends up being confusing.

How about something building like, which is much easier to read IMO:
```
Encountered error(s) while parsing foo.rb: 
- expected a `)` to close the parameters (at 1:8).
- cannot parse the expression (at 1:8). 
- expected an `end` to close the `def` statement (at 1:8).
```

paracycle: I am not sure if it is a requirement for the `message` to be a single line. After all, the docs for `detailed_message` do talk about decorating the _first line_ of the exception message which implies that it might be expecting more than 1 line: https://ruby-doc.org/3.3.0/Exception.html#method-i-detailed_message
> It may add the class name of the exception to the end of the first line.

There is another option at your disposal as well, though. You could store the formatted errors in your custom exception, and do the formatting at the site that's doing the handling. You could even encapsulate the formatting in the `to_s` of the exception class.

paracycle (approved):

PR #538, opened 2024-04-16
Fix removing first method in class

skipkayhil:
when following by a multi-line node.

Previously, trying to remove the first method in a module/class when there was another method after it would also remove all but the last line of the following method as well. The problem can be seen below:

```
module Foo
  def foo
  end
          # after.location.start_line - 1
  def bar # after.location.end_line - 1
  end
end
```

To account for blank lines, the end_line of the node to remove needs to be adjusted to include all lines upto the _start_ of the next node, whereas the code was previously including all lines upto the _end_ of the next node.

### DIFF
--- a/lib/spoom/deadcode/remover.rb
+++ b/lib/spoom/deadcode/remover.rb
@@ -312,7 +312,7 @@ module Spoom
           # Adjust the lines to remove to include following blank lines
           after = context.next_node
           if before.nil? && after && after.location.start_line > end_line + 1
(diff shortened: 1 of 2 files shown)


skipkayhil:
I had an additional thought that many of the tests in this file use single line method/class definition, which could hide other bugs where a node's `start_line == end_line`. I wonder if some of them should be changed to use multi-line method/class definitions for better coverage of those cases.

PR #543, opened 2024-05-02
Fix removal of def nodes with a rescue

Morriar:
When the def node contains a `rescue` we were matching against the `BeginNode` location instead of the `DefNode`.

To this this we can stop the descendant lookup as soon as we find a node that as both the right location and kind.

### DIFF
--- a/test/spoom/deadcode/remover_test.rb
+++ b/test/spoom/deadcode/remover_test.rb
@@ -1027,6 +1027,26 @@ module Spoom
         RB
       end
 
+      def test_deadcode_remover_removes_method_with_rescue
+        res = remove(<<~RB, "foo")
+          def bar; end
+
+          def foo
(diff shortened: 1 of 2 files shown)


test/spoom/deadcode/remover_test.rb
```diff
@@ -1027,6 +1027,26 @@ def bar; end
         RB
       end
 
+      def test_deadcode_remover_removes_method_with_rescue
```
andyw8, line 1030: (nit) Not specific to this PR, but it seems a redundant to start each test name with `test_deadcode_remover_` when we're already inside `DeadCode::RemoverTest`.

Morriar: I'm open to a renaming PR 😂

PR #561, opened 2024-06-18
Compute class and module hierarchy

Morriar:
Now that we have a complete model (#557) and a poset (#558) we can compute the full class/module hierarchy.

This will allow deadcode plugins to reason about the whole ancestor chain rather than just the parent class.

### DIFF
--- a/test/spoom/model/model_test.rb
+++ b/test/spoom/model/model_test.rb
@@ -1,0 +1,66 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module Spoom
+  class Model
+    class ModelTest < Minitest::Test
+      extend T::Sig
+
+      def test_resolve_symbol
+        model = model(<<~RB)
+          module Foo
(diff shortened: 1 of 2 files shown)


test/spoom/model/model_test.rb
```diff
@@ -0,0 +1,119 @@
...
+        model = model(<<~RB)
+          class A; end
+          class B < A; end
+          class C < B; end
+          class D; end
+        RB
+
+        assert_equal(
+          ["B", "C"],
+          model.subtypes(model["A"]).map(&:full_name).sort,
```
KaanOzkan, line 62: Should we not test the order? Why do we sort?

KaanOzkan (approved):

test/spoom/model/model_test.rb
Morriar: Good question: The order is not relevant at this point since we do not do any kind of linearization (yet).

PR #562, opened 2024-06-19
Make dead code indexing use a `Model` to index classes

Morriar:
This is the first PR on a sequence moving the dead code analysis use the `Model` introduced in #557 rather than its custom indexer.

In this PR we migrate the indexing of classes:
* Make `Deadcode::Index` take a `Model` (bc1852074a4826acf12d701ee0186515ef8999ba)
* Remove classes indexing from `Deadcode::Indexer` (9433fda1d7f1a6655d9f872af4f579bc30fa9f5e)
* Make the `Deadcode::Plugins` take `Model::SymbolDef` when calling `on_define_class` (635977905fa69baed567f1e04ade5456dde2ade9)

There is no behavior change: no tests are changed.

### DIFF
--- a/lib/spoom/deadcode/plugins/base.rb
+++ b/lib/spoom/deadcode/plugins/base.rb
@@ -159,26 +159,26 @@ module Spoom
         #
         # ~~~rb
         # class MyPlugin < Spoom::Deadcode::Plugins::Base
-        #   def on_define_class(indexer, definition)
-        #     definition.ignored! if definition.name == "Foo"
+        #   def on_define_class(symbol_def, definition)
+        #     definition.ignored! if symbol_def.name == "Foo"
         #   end
         # end
         # ~~~
-        sig { params(indexer: Indexer, definition: Definition).void }
-        def on_define_class(indexer, definition)
+        sig { params(symbol_def: Model::Class, definition: Definition).void }
+        def on_define_class(symbol_def, definition)
           # no-op
         end
 
         # Do not override this method, use `on_define_class` instead.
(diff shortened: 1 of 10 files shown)


lib/spoom/deadcode/plugins/base.rb
```diff
@@ -159,26 +159,26 @@ def internal_on_define_accessor(indexer, definition)
...
-        #   def on_define_class(indexer, definition)
-        #     definition.ignored! if definition.name == "Foo"
+        #   def on_define_class(symbol_def, definition)
+        #     definition.ignored! if symbol_def.name == "Foo"
         #   end
         # end
         # ~~~
-        sig { params(indexer: Indexer, definition: Definition).void }
-        def on_define_class(indexer, definition)
+        sig { params(symbol_def: Model::Class, definition: Definition).void }
```
KaanOzkan, line 167: Should this be abstract?

KaanOzkan (approved):

lib/spoom/deadcode/plugins/base.rb
Morriar: We don't want to crash when calling `on_define_class` to a plugin that do nothing with it, we just want it to no-op.

We _could_refactor this to have multiple interfaces:

```rb
module ClassPlugin
  interface!

  sig { abstract.void }
  def on_define_class; end
```

and then let the plugins cherry pick the interfaces they want to implement.

But in practice I just rely on YJIT to optimize this empty calls away.

egiurleo (approved):

PR #566, opened 2024-06-19
Make dead code indexing use a Model to index accessors

Morriar:
Follow-up on #565/

This time, we switch the indexing of accessors to Model:

* Remove accessors indexing from Deadcode::Indexer + some cleaning (0a5d0dffd12059e817ef6fb8ec1958da91d7be20)
Make the Deadcode::Plugins take Model::SymbolDef when calling `on_define_accessor` (fea73f3b3fd3b107c46bdb28ed012a10514355bb)

There is no behavior change. 

### DIFF
--- a/lib/spoom/deadcode/indexer.rb
+++ b/lib/spoom/deadcode/indexer.rb
@@ -17,29 +17,12 @@ module Spoom
         super()
 
         @path = path
-        @file_name = T.let(File.basename(path), String)
         @source = source
         @index = index
         @plugins = plugins
-        @previous_node = T.let(nil, T.nilable(Prism::Node))
-        @names_nesting = T.let([], T::Array[String])
-        @nodes_nesting = T.let([], T::Array[Prism::Node])
-        @in_const_field = T.let(false, T::Boolean)
-        @in_opassign = T.let(false, T::Boolean)
-        @in_symbol_literal = T.let(false, T::Boolean)
       end
 
       # Visit
-
-      sig { override.params(node: T.nilable(Prism::Node)).void }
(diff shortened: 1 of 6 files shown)


lib/spoom/deadcode/indexer.rb
```diff
@@ -330,62 +217,6 @@ def reference_method(name, node)
         @index.reference(Reference.new(name: name, kind: Reference::Kind::Method, location: node_location(node)))
       end
 
-      # Context
-
-      sig { returns(Prism::Node) }
-      def current_node
```
KaanOzkan, line 336: Were these methods unused? There are some definitions that are removed without any callsites being removed.

lib/spoom/deadcode/indexer.rb
```diff
@@ -17,30 +17,13 @@ def initialize(path, source, index, plugins: [])
         super()
 
         @path = path
-        @file_name = T.let(File.basename(path), String)
         @source = source
         @index = index
         @plugins = plugins
-        @previous_node = T.let(nil, T.nilable(Prism::Node))
-        @names_nesting = T.let([], T::Array[String])
```
KaanOzkan, line 25: Why don't we have to keep track of nesting anymore?

lib/spoom/deadcode/indexer.rb
Morriar: They were helpers for the plugins to access more context around the node it was analyzing. All this knowledge is now moved inside the `Model` so plugins can access it more easily.

egiurleo (approved):

PR #579, opened 2024-07-04
Improve deadcode plugin for Minitest

Morriar:
* Make plugin ignore classes that inherit from `Minitest::Test` instead of those ending with `Test`
* Modified the `on_define_method` method to ignore methods starting with `test_` or included in the `MINITEST_METHODS` set only when inside a `Minitest::Test` subclass.
* Define `on_send` to mark methods used in `assert_predicate` and `refute_predicate` as alive.

### DIFF
--- a/lib/spoom/deadcode/plugins/minitest.rb
+++ b/lib/spoom/deadcode/plugins/minitest.rb
@@ -7,7 +7,7 @@ module Spoom
       class Minitest < Base
         extend T::Sig
 
-        ignore_classes_named(/Test$/)
+        ignore_classes_inheriting_from("Minitest::Test")
 
         ignore_methods_named(
           "after_all",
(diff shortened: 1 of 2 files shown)


lib/spoom/deadcode/plugins/minitest.rb
```diff
@@ -7,21 +7,39 @@ module Plugins
...
+            "teardown",
+          ]),
+          T::Set[String],
         )
 
         sig { override.params(definition: Model::Method).void }
         def on_define_method(definition)
-          file = definition.location.file
-          @index.ignore(definition) if file.match?(%r{test/.*test\.rb$}) && definition.name.match?(/^test_/)
+          return unless definition.name =~ /^test_/ || MINITEST_METHODS.include?(definition.name)
```
amomchilov, line 26: regex is overkill here

```suggestion
          return unless definition.name.start_with?("test_") || MINITEST_METHODS.include?(definition.name)
```

amomchilov (approved):

PR #585, opened 2024-07-12
Add Tests and Fix Comparison Bug in `Spoom::Location`

alexcrocha:
## Summary
This PR addresses the lack of tests for the `Spoom::Location` class by introducing test coverage for its methods.

## Details
- **Test Coverage**: Added tests to ensure all methods in `Spoom::Location` function as expected.

- **Bug Fix**: During the creation of these tests, a logical bug was identified in the implementation of the `<=>` method. The current implementation only compares strings, which leads to incorrect comparisons for integers (e.g., `"11" < "2"` evaluates as true, but `2 < 11` is the correct comparison).

### DIFF
--- a/lib/spoom/location.rb
+++ b/lib/spoom/location.rb
@@ -73,7 +73,8 @@ module Spoom
     def <=>(other)
       return unless Location === other
 
-      to_s <=> other.to_s
+      [file, start_line, start_column, end_line, end_column] <=>
+        [other.file, other.start_line, other.start_column, other.end_line, other.end_column]
     end
 
     sig { returns(String) }
--- a/test/spoom/location_test.rb
+++ b/test/spoom/location_test.rb
@@ -1,0 +1,92 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module Spoom
+  module Sorbet
+    class LocationTest < Minitest::Test
+      def test_from_string
+        location = Location.from_string("foo.rb:1:2-3:4")
+        assert_equal("foo.rb", location.file)
+        assert_equal(1, location.start_line)
+        assert_equal(2, location.start_column)
+        assert_equal(3, location.end_line)
+        assert_equal(4, location.end_column)
+      end
+
+      def test_raises_if_location_string_has_missing_components
+        assert_raises(Location::LocationError) do
+          Location.from_string("foo.rb:1:2-3")
+        end
+
+        assert_raises(Location::LocationError) do
+          Location.from_string("foo.rb:1:2")
+        end
+
+        assert_raises(Location::LocationError) do
+          Location.from_string("foo.rb:1")
+        end
+
+        assert_raises(Location::LocationError) do
+          Location.from_string("foo.rb")
+        end
+      end
+
+      def test_include
+        location1 = Location.new("foo.rb", 1, 2, 3, 4)
+        location2 = Location.new("foo.rb", 1, 2, 3, 4)
+        assert(location1.include?(location2))
+
+        location3 = Location.new("foo.rb", 1, 2, 3, 5)
+        refute(location1.include?(location3))
+        assert(location3.include?(location1))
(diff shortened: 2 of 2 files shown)


test/spoom/location_test.rb
```diff
@@ -0,0 +1,100 @@
...
+          Location.from_string("foo.rb")
+        end
+      end
+
+      def test_include
+        location1 = Location.new("foo.rb", 1, 2, 3, 4)
+        location2 = Location.new("foo.rb", 1, 2, 3, 4)
+        assert(location1.include?(location2))
+
+        location3 = Location.new("foo.rb", 1, 2, 3, 5)
```
andyw8, line 44: The test could be a little more readable if we give descriptive names to the variables, e.g. `location_with_higher_column`.

test/spoom/location_test.rb
```diff
@@ -0,0 +1,100 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+require_relative "../../lib/spoom/location.rb"
+
+require "stringio"
```
andyw8, line 7: This doesn't seem needed?

lib/spoom/location.rb
```diff
@@ -73,7 +73,19 @@ def include?(other)
     def <=>(other)
       return unless Location === other
 
-      to_s <=> other.to_s
+      res = file <=> other.file
```
andyw8, line 76:
```
[file, start_line, start_column, end_line, end_column] <=>
        [other.file, other.start_line, other.start_column, other.end_line, other.end_column]
```

test/spoom/location_test.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      def test_from_string
+        location = Location.from_string("foo.rb:1:2-3:4")
+        assert_equal("foo.rb", location.file)
+        assert_equal(1, location.start_line)
+        assert_equal(2, location.start_column)
+        assert_equal(3, location.end_line)
+        assert_equal(4, location.end_column)
+
+        assert_raises(Location::LocationError) do
```
andyw8, line 22: (nit) I'd be inclined split these into a separate test so it can be given a meaningful name, e.g. `test_raises_if_location_string_has_missing_components`.

andyw8:
Thanks for improving the coverage!

test/spoom/location_test.rb
alexcrocha: So much better! I ended up dropping `location_with_` as it was getting too long and making it hard to read. I wonder if the variable names still make sense without?

test/spoom/location_test.rb
alexcrocha: 🙏 Thanks for catching. Removing.

lib/spoom/location.rb
alexcrocha: I like the suggestion. Easier to read and maintain this way.

> But on further thought, aren't end_line and end_column irrelevant here?

I am under the assumption we still want to compare if the the `end_line` or `end_column` are different. But I may be missing the big picture. Could you elaborate?

test/spoom/location_test.rb
alexcrocha: Makes sense 👍

lib/spoom/location.rb
andyw8: I suspect that when we sort or compare locations, we're really interested in where it _begins_, i.e. the `file`, `start_line` and `start_colum`. May be worth looking into.

test/spoom/location_test.rb
alexcrocha: Just to make sure we're all on the same page, should I revert to using `location1`, `location2` naming, or keep the updated descriptive variable names?

andyw8: You can put it back to `location1` etc.

Morriar (approved):
Thanks!

PR #592, opened 2024-07-29
Do not mark classes that inherit from `ActiveJob::TestCase` as dead

nvasilevski:
In https://github.com/Shopify/spoom/pull/579 we stopped ignoring class names that end with `Test` which caused test classes that inherit from `ActiveJob::TestCase` to be indexed as dead. It's not an issue for classes inherited from `ActiveSupport::TestCase` because we have a plugin that ignores those:
https://github.com/Shopify/spoom/blob/3e8ae2092aad7561476925c47a15bdaaeedaa1b5/lib/spoom/deadcode/plugins/active_support.rb#L8

We should do the same for `ActiveJob::TestCase`. Since we already have `ActiveJob` plugin I add an ignore macro to the plugin.

### DIFF
--- a/lib/spoom/deadcode/plugins/active_job.rb
+++ b/lib/spoom/deadcode/plugins/active_job.rb
@@ -5,6 +5,7 @@ module Spoom
   module Deadcode
     module Plugins
       class ActiveJob < Base
+        ignore_classes_inheriting_from("ActiveJob::TestCase")
         ignore_classes_named("ApplicationJob")
         ignore_methods_named("perform", "build_enumerator", "each_iteration")
       end
--- a/test/spoom/deadcode/plugins/active_job_test.rb
+++ b/test/spoom/deadcode/plugins/active_job_test.rb
@@ -49,6 +49,15 @@ module Spoom
           )
         end
 
+        def test_ignores_activejob_testcase_class_definition
+          @project.write!("test/foo_test.rb", <<~RB)
+            class FooTest < ActiveJob::TestCase
+            end
+          RB
+
+          assert_ignored(index_with_plugins, "FooTest")
+        end
+
         private
 
         sig { returns(Deadcode::Index) }


lib/spoom/deadcode/plugins/active_job.rb
```diff
@@ -5,6 +5,7 @@ module Spoom
   module Deadcode
     module Plugins
       class ActiveJob < Base
+        ignore_classes_inheriting_from("ActiveJob::TestCase")
```
Morriar, line 8: Isn't `ActiveJob::TestCase` a subclass of `ActiveSupport::TestCase` (https://github.com/rails/rails/blob/main/activejob/lib/active_job/test_case.rb#L6)? The `ActiveSupport` plugin should be catching this so there is another kind of bug at play here. I don't think this is the right fix.

nvasilevski: >Isn't ActiveJob::TestCase a subclass of ActiveSupport::TestCase

It definitely is but since it wasn't working I just assumed that "for reasons" we only look at the direct parent without trying to analyze the whole inheritance chain.

But if we expect `ignore_classes_inheriting_from` to apply to the whole chain then I totally agree we should be looking into keeping only `ActiveSupport::TestCase` as ignored parent and fixing the lookup instead

>I wonder what would be the impact of also adding .rbi files to the Model

I can allocate some time to do that but since we are experiencing an issue right now and current version has a bug how would you feel about either merging this PR with explicit ignore of `ActiveJob::TestCase` or perhaps temporarily removing changes to `Minitest` plugin (bringing back ignore of `*Test` classes

Morriar: I think we can merge this. Though, since you know Rails can you ensure that we not missing other subclasses of `ActiveSupport::TestCase`?

I'll handle the RBI indexing exploration 👍

Morriar (approved):

nvasilevski:
Since `ActiveJob` is not the only framework with its own subclass we are going to temporarily bring back the exclusion of `*Test` classes until we find a proper solution which will allow only `ActiveSupport` parent to be ignored
image

PR #593, opened 2024-07-29
Ignore `*Test` named classes again

nvasilevski:
Partially reverts https://github.com/Shopify/spoom/pull/579

Not ignoring classes named `*Test` end up resulting in a bug where test classes that inherit from Rails' framework specific test class like `ActiveJob::TestCase` started to be considered dead. https://github.com/Shopify/spoom/pull/592
 
All of the framework test cases inherit from `ActiveSupport::TestCase` so ultimately it should be enough to have it ignored
https://github.com/Shopify/spoom/blob/d4ccf209f2f244965be77d3644fc646ea26ff613/lib/spoom/deadcode/plugins/active_support.rb#L8

should suffice but at the moment spoom doesn't see the `AJ::TestCase < AS::TestCase` inheritance chain so until we get it fixed we are bringing the `*Test` ignore rule back

### DIFF
--- a/lib/spoom/deadcode/plugins/minitest.rb
+++ b/lib/spoom/deadcode/plugins/minitest.rb
@@ -7,6 +7,7 @@ module Spoom
       class Minitest < Base
         extend T::Sig
 
+        ignore_classes_named(/Test$/)
         ignore_classes_inheriting_from("Minitest::Test")
 
         MINITEST_METHODS = T.let(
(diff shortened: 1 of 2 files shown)


Morriar (approved):

Morriar:
@nvasilevski do you mean creating a new Spoom release with this fix please? 🙏

PR #603, opened 2024-09-02
Update hint to remove deprecated command

aericson:
Update hint after running `srb bump` to not reference the deprecated command.


egiurleo (approved):
Thanks for catching that!

PR #608, opened 2024-09-30
Return Gem::Version from Bundle#gem_version_from_gemfile_lock

KaanOzkan:
Started using this in `rbi-central` and returning a `Gem::Version` is closer to what a user would expect.

https://github.com/Shopify/rbi-central/pull/286#discussion_r1781353676


Morriar (approved):
Thanks! ❤️

PR #624, opened 2024-12-18
Add `spoom srb sigs strip` command

Morriar:
Add a command to strip Sorbet signatures from files.

I refactored the signature translation code to reuse some logic, this PR is easier to review commit by commit.

cc. @dirceu

### DIFF
--- a/lib/spoom/cli/srb/sigs.rb
+++ b/lib/spoom/cli/srb/sigs.rb
@@ -7,6 +7,7 @@ module Spoom
   module Cli
     module Srb
       class Sigs < Thor
+        extend T::Sig
         include Helper
 
         desc "translate", "Translate signatures from/to RBI and RBS"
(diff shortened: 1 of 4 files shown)


lib/spoom/cli/srb/sigs.rb
```diff
@@ -65,6 +51,28 @@ def collect_files(paths)
...
           end
+
+          def transform_files(files, &block)
+            transformed_count = 0
+
+            files.each do |file|
+              contents = File.read(file)
+
+              if contents.lines.first&.start_with?("# encoding:")
+                encoding = T.must(contents.lines.first).gsub(/^#\s*encoding:\s*/, "").strip
```
KaanOzkan, line 71: Since we do this for each file it might be worth perf wise to store `contents.lines.first` in a variable and above and reuse it below.

KaanOzkan (approved):

lib/spoom/cli/srb/sigs.rb
Morriar: I'm confused, can you explain what you mean?

PR #632, opened 2025-01-23
Remove duplicate case statement

KaanOzkan:


Morriar (approved):

andyw8:
🤔 Let's enable that RuboCop check? https://github.com/Shopify/ruby-style-guide/blob/main/rubocop.yml#L132

PR #695, opened 2025-03-10
Add RBS to RBI translation

Morriar:
This PR introduces a new command to translate RBS comments back into Sorbet `sig{}` blocks.

Given this file:

```rb
#: -> void
def foo; end
```

Running `spoom srb sigs translate --from rbi --to rbs` will rewrite the file into:

```rb
sig { void }
def foo; end
```

### DIFF
--- a/lib/spoom/cli/srb/sigs.rb
+++ b/lib/spoom/cli/srb/sigs.rb
@@ -10,18 +10,31 @@ module Spoom
         include Helper
 
         desc "translate", "Translate signatures from/to RBI and RBS"
-        option :from, type: :string, aliases: :f, desc: "From format", enum: ["rbi"], default: "rbi"
-        option :to, type: :string, aliases: :t, desc: "To format", enum: ["rbs"], default: "rbs"
+        option :from, type: :string, aliases: :f, desc: "From format", enum: ["rbi", "rbs"], default: "rbi"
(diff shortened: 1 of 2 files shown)


lib/spoom/sorbet/sigs.rb
```diff
@@ -34,7 +35,28 @@ def rbi_to_rbs(ruby_contents)
...
+            scanner = Scanner.new(ruby_contents)
+            start_index = scanner.find_char_position(
+              T.must(rbs_comment.loc&.begin_line&.pred),
+              T.must(rbs_comment.loc).begin_column,
+            )
+            end_index = scanner.find_char_position(
+              rbs_comment.loc&.end_line&.pred,
+              T.must(rbs_comment.loc).end_column,
+            )
+            ruby_contents[start_index...end_index] = RBSToRBITranslator.translate(rbs_comment, node)
```
KaanOzkan, line 59: FYI I used `..` in tapioca. Can't remember exactly why but I think I was missing a new line otherwise. Tests here look good so I don't think we need it but I'm intrigued now.

KaanOzkan (approved):

lib/spoom/sorbet/sigs.rb
Morriar: That's why I'm using `...` to exclude the end index which is the index of the `\n`, since I `.strip` the signature I'm creating.

Maybe you were not stripping the sig, meaning you had an extra new line after the sig and needed to remove the `\n` from the original one?

PR #697, opened 2025-03-11
Add `spoom srb sigs export` command

Morriar:
Add a command to export all the RBS signatures from the project as a single `.rbi` file:

```
$ spoom srb sigs export
```

This command will write a file at `rbi/spoom.rbi` containing all the sigs from the project translated as `sig{}` blocks.

It also comes with an option to check this file is up to date so we can run it on CI:

```
$ spoom srb sigs export --check-sync
```

This command won't actually write the file, it just checks there is not difference between the exported one and the one under `rbi/spoom.rbi`.


lib/spoom/cli/srb/sigs.rb
```diff
@@ -53,6 +53,118 @@ def strip(*paths)
...
+            say_error("No entry point found at `#{entry_point}`")
+            exit(1)
+          end
+
+          say("Injecting `extend T::Sig` to `#{entry_point}`...")
+          copy_context.write!(entry_point, <<~RB)
+            require "sorbet-runtime"
+
+            class Module; include T::Sig; end
+            extend T::Sig
```
KaanOzkan, line 107: Do we need this extend?

lib/spoom/cli/srb/sigs.rb
```diff
@@ -53,6 +53,118 @@ def strip(*paths)
...
+        # This command will use Tapioca to generate a `.rbi` file that contains the signatures of all the files listed
+        # in the gemspec.
+        desc "export", "Export gem files signatures"
+        option :gemspec, type: :string, desc: "Path to the gemspec file", optional: true, default: nil
+        option :check_sync, type: :boolean, desc: "Check the generated RBI is up to date", default: false
+        def export(output_path = nil)
+          gemspec = options[:gemspec]
+
+          unless gemspec
+            say("Locating gemspec file...")
```
KaanOzkan, line 67: Not a big deal but we could clean up the nested `unless` if we're okay with getting rid of this say
`gemspec = options[:gemspec] || Dir.glob("*.gemspec").first`

lib/spoom/cli/srb/sigs.rb
```diff
@@ -53,6 +53,118 @@ def strip(*paths)
           say("Stripped signatures from `#{transformed_files}` file#{transformed_files == 1 ? "" : "s"}.")
         end
 
+        # Extract signatures from gem files and save them to the output file
```
KaanOzkan, line 56:
```suggestion
        # Extract signatures from gem's files and save them to the output file
```
I think it'll help with clarity since we don't want them to think `Gemfile`.

PR #725, opened 2025-04-29
Upgrade development Ruby to 3.4.3

vinistock:
Let's use the latest Ruby for development. I also added the 3.4 series to our CI.


.github/workflows/ruby.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        ruby: ["3.1", "3.2", "3.3"]
+        ruby: ["3.1", "3.2", "3.3", "3.4"]
```
Morriar, line 18: 3.1 is EOL since 2025-03-26, I think we can remove it from the matrix

Morriar (approved):

PR #734, opened 2025-05-15
Use `Source::Rewriter` for RBS sigs and assertions translation.

Morriar:
As suggested by @vinistock in https://github.com/Shopify/spoom/pull/733#issuecomment-2883880667.

### DIFF
--- a/rbi/spoom.rbi
+++ b/rbi/spoom.rbi
@@ -2556,115 +2556,6 @@ end
 
 Spoom::SPOOM_PATH = T.let(T.unsafe(nil), String)
 module Spoom::Sorbet; end
-
-class Spoom::Sorbet::Assertions
-  class << self
-    sig { params(ruby_contents: ::String, file: ::String).returns(::String) }
-    def rbi_to_rbs(ruby_contents, file:); end
-
-    private
-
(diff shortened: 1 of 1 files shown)


lib/spoom/sorbet/translate/translator.rb
```diff
@@ -0,0 +1,62 @@
...
+      # @abstract
+      class Translator < Spoom::Visitor
+        #: (String, file: String) -> void
+        def initialize(ruby_contents, file:)
+          super()
+
+          @file = file #: String
+
+          @original_encoding = ruby_contents.encoding #: Encoding
+          @ruby_contents = if ruby_contents.encoding == "UTF-8"
```
vinistock, line 16:
```suggestion
          @ruby_contents = if @original_encoding == "UTF-8"
```

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -0,0 +1,232 @@
+# typed: true
+# frozen_string_literal: true
+
+require "test_helper"
+
+module Spoom
+  module Sorbet
+    module Translate
+      class RBSCommentsToSorbetSigsTest < Minitest::Test
```
vinistock, line 9: Do we have a test for multi line RBS signatures? Should we add one?

PR #741, opened 2025-05-20
Collect metrics about Sorbet and RBS usage

Morriar:
This provides a visitor that collects metrics about how Sorbet and RBS are used in a codebase:

```rb
context = Context.mktmp!
context.write!("foo.rb", <<~RUBY)
  x = 1 #: Integer
  x = 1 #: as !nil
  x = 1 #: as untyped
  x = 1 #: as Integer
RUBY

metrics = Spoom::Sorbet::Metrics.collect_code_metrics(Dir["#{context.absolute_path}/**/*.rb"])

assert_equal(4, metrics["rbs_assertions"])
assert_equal(1, metrics["rbs_let"])
assert_equal(1, metrics["rbs_must"])
assert_equal(1, metrics["rbs_unsafe"])
assert_equal(1, metrics["rbs_cast"])
```

It can also be ran from the CLI with `spoom srb metrics <path>`:

image


lib/spoom/counters.rb
```diff
@@ -0,0 +1,30 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Spoom
+  class Counters
+    #: Hash[String, Integer]
+    attr_reader :counters
```
vinistock, line 7: I wonder if we should just inherit from `Hash`, so that you can invoke `sort_by` directly on the object.

lib/spoom/cli/srb/metrics.rb
```diff
@@ -0,0 +1,68 @@
...
+
+          ["methods", "accessors"].each do |key|
+            value = metrics[key]
+            next if value == 0
+
+            say("\n#{key.capitalize}: `#{value}`")
+            ["without_sig", "with_srb_sig", "with_rbs_sig"].each do |subkey|
+              say(" * #{subkey.gsub("_", " ")}: `#{metrics["#{key}_#{subkey}"]}`")
+            end
+          end
```
vinistock, line 46: This feels like we're trying to represent nested data with a flat structure. Would it make sense to have nested hashes for the metrics and then replace all of this code with just a pretty print of the hash?

Morriar: I'm not sure.

The grouping only happens because I want to display the metrics in a cute way.

For most of other uses I have (CodeDB or pasting into a spreadsheet) it's not that important.

I like the simplicity, versatility and extensibility of the flat counters.

PR #748, opened 2025-06-03
Mark ActiveJob::Callbacks as alive

Morriar:
See https://api.rubyonrails.org/v7.1/classes/ActiveJob/Callbacks/ClassMethods.html.

### DIFF
--- a/lib/spoom/deadcode/plugins/active_job.rb
+++ b/lib/spoom/deadcode/plugins/active_job.rb
@@ -7,6 +7,25 @@ module Spoom
       class ActiveJob < Base
         ignore_classes_named("ApplicationJob")
         ignore_methods_named("perform", "build_enumerator", "each_iteration")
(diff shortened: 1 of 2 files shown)


lib/spoom/deadcode/plugins/active_job.rb
```diff
@@ -7,6 +7,25 @@ module Plugins
...
+          "before_enqueue",
+          "before_perform",
+        ].freeze
+
+        # @override
+        #: (Send send) -> void
+        def on_send(send)
+          return unless send.recv.nil? && CALLBACKS.include?(send.name)
+
+          send.each_arg(Prism::SymbolNode) do |arg|
```
vinistock, line 25: I think it probably accepts strings too if you want to be exhaustive

vinistock (approved):

Issue #752, opened 2025-06-06
Sig translation should output multiline RBS comments when needed


KaanOzkan:
Right now all RBS comments are outputted on a single line. `rbi` is [capable](https://github.com/Shopify/rbi/blob/9526a25239775209e11286d4ab7f2ccad099abd6/lib/rbi/rbs_printer.rb#L477) of printing multiline RBS signatures but it doesn't contain the necessary `#|` prefix. It can be improved in `rbi` and used in `spoom` after a certain character limit.

Morriar:
Fix https://github.com/Shopify/spoom/pull/761

PR #753, opened 2025-06-06
RBS sig translate should generate multiline when the line is too long

jesse-shopify:
- Possible solution to https://github.com/Shopify/spoom/issues/752

We coded this to see if it could be handled in spoom. Discussion was whether this should be in `rbi` instead. This might have some value as a reference.


KaanOzkan:
Done in https://github.com/Shopify/spoom/commit/8926c143dd12a681368b07f6f821638c566165fa

PR #754, opened 2025-06-06
Tests for simplifying and normalizing types in sigs during translation

jesse-shopify:
Tests for normalization / simplification / flattening of
- nested `T.nilable`
- nested `T.any`
- nested `T.all`

Requires RBI with https://github.com/Shopify/rbi/pull/479/

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -52,10 +52,9 @@ GEM
     racc (1.8.1)
     rainbow (3.1.1)
     rake (13.3.0)
-    rbi (0.3.3)
+    rbi (0.3.5)
       prism (~> 1.0)
       rbs (>= 3.4.4)
-      sorbet-runtime (>= 0.5.9204)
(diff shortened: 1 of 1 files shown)


test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
```diff
@@ -288,6 +288,54 @@ def foo
           RBS
         end
 
+        def test_translate_to_rbs_with_nested_nilable_param
+          contents = <<~RB
+            Class.new do
+              sig { params(x: T.nilable(T.nilable(Integer))).void }
```
amomchilov, line 294: Is our logic for this purpose-built for RBS? This kind of normalization would be useful more generally. Would be nice to do: `sig -> normalized sig -> RBS`

Morriar (approved):

PR #755, opened 2025-06-09
Clear last_sigs when returning early on abstract

Morriar:
Before this fix, when encountering a signature with `abstract` we were not clearing the list, meaning that subsequent signatures were not translated:

```rb
sig { abstract.void }
def foo; end

sig { void } # list still contains the abstract one and will return early
def bar; end
```

After this change, the second signature is properly translated:

```rb
sig { abstract.void }
def foo; end

#: -> void
def bar; end
```

### DIFF
--- a/rbi/spoom.rbi
+++ b/rbi/spoom.rbi
@@ -2995,6 +2995,9 @@ class Spoom::Sorbet::Translate::SorbetSigsToRBSComments < ::Spoom::Sorbet::Trans
   sig { params(node: ::Prism::ConstantWriteNode).returns(::String) }
   def build_type_member_string(node); end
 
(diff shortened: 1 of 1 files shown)


lib/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments.rb
```diff
@@ -336,6 +334,13 @@ def delete_extend_t_generics
 
           @extend_t_generics.clear
         end
+
+        #: -> Array[[Prism::CallNode, RBI::Sig]]
+        def collect_last_sigs
```
KaanOzkan, line 339: Maybe a comment here is useful that mentions resetting, it's a bit strange looking.

KaanOzkan (approved):

PR #765, opened 2025-06-13
Allow translation of abstract methods

Morriar:
Goes from this:

```rb
class Foo
  sig { abstract.void }
  def foo; end
end
```

to this:

```rb
class Foo
  # @abstract
  #: -> void
  def foo; end
end
```

Should we also insert the `raise` in the body while at it?

### DIFF
--- a/test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
+++ b/test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
@@ -74,19 +74,29 @@ module Spoom
 
         def test_does_not_translate_to_rbs_abstract_methods
           contents = <<~RB
-            sig { abstract.void }
-            def foo; end
+            class Foo
+              sig { abstract.void }
+              def foo; end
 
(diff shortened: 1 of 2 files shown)


paracycle (commented):
I think we should insert:
```ruby
raise NotImplementedError, "Abstract method called"
```
automatically if Sorbet is then later going to require it anyway.

juharris (approved):
Nice thanks! We were just noticing this on our team.

Morriar:
Also added an autocorrect on the body to insert the `raise` 👍

test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
```diff
@@ -74,19 +74,29 @@ def a
 
         def test_does_not_translate_to_rbs_abstract_methods
```
KaanOzkan, line 75: Test name needs updating

KaanOzkan (approved):

jesse-shopify (approved):

PR #771, opened 2025-06-18
Make translation of each kind of T assertions optional

Morriar:
By default we translate everything.

It can be enabled/disabled separately:

```
$ spoom srb assertions translate --translate-t-let --no-translate-t-must --no-translate-t-cast --no-translate-t-bind --no-translate-t-unsafe
```

### DIFF
--- a/rbi/spoom.rbi
+++ b/rbi/spoom.rbi
@@ -2842,8 +2842,18 @@ module Spoom::Sorbet::Translate
     sig { params(ruby_contents: ::String, file: ::String, max_line_length: T.nilable(::Integer)).returns(::String) }
     def rbs_comments_to_sorbet_sigs(ruby_contents, file:, max_line_length: T.unsafe(nil)); end
 
-    sig { params(ruby_contents: ::String, file: ::String).returns(::String) }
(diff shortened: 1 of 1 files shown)


lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
```diff
@@ -120,12 +147,16 @@ def t_annotation?(node)
           return false unless t?(node.receiver)
 
           case node.name
-          when :let, :cast
-            return node.arguments&.arguments&.size == 2
+          when :let
```
vinistock, line 150: It feels a little odd to have these options checks as part of this method. This method intuitively checks if something _is_ an annotation. The fact that the translate option is turned off doesn't make the thing not be an annotation anymore.

Would it require too much work to move this out? Or maybe we could have another method that calls this one called `should_translate?`. What do you think?

Morriar: The thing is I don't want to check twice on the name of the send.

I renamed the method `translatable_annotation?` instead.

vinistock (approved):

PR #773, opened 2025-06-27
Add a test showing RBS signatures are being deleted by the deadcode remover

Morriar:
@shioyama it looks like it works. Do you have an example where it didn't?

### DIFF
--- a/test/spoom/deadcode/remover_test.rb
+++ b/test/spoom/deadcode/remover_test.rb
@@ -965,6 +965,29 @@ module Spoom
         RB
       end
 
+      def test_removes_node_rbs_comment
+        res = remove(<<~RB, "bar")
+          class Foo
+            def foo; end
+
+            #: -> void
+            def bar
+              something
+            end
+
+            def baz; end
+          end
+        RB
+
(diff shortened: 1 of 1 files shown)


KaanOzkan (commented):
I think we can close since we have https://github.com/Shopify/spoom/blob/747c18a50a050abf061ef585be582638ff711338/test/spoom/deadcode/remover_test.rb#L965-L988

test/spoom/deadcode/remover_test.rb
```diff
@@ -965,6 +965,29 @@ def baz; end
         RB
       end
 
+      def test_removes_node_rbs_comment
```
Morriar: I wonder if we should either close this as covered by `test_removes_node_with_comments`, or make the RBS-specific regression more explicit. Right now `#:` is still just an attached Prism comment for the remover, so this exercises the same path as the existing comments test.

If we want to keep a dedicated RBS regression, maybe use a case that documents why RBS comments are special here, like a multi-line signature / annotation block that would be easy to leave behind accidentally.

Morriar:
Very stale

PR #780, opened 2025-07-21
Do not associate a RBS comment to a node after a blank line

Morriar:
Before this PR, the following example would fail because we tried to use the type alias as the generic signature of the class:

```rb
#: type foo = Integer

class Bar; end
```

### DIFF
--- a/lib/spoom/rbs.rb
+++ b/lib/spoom/rbs.rb
@@ -75,8 +75,16 @@ module Spoom
 
         continuation_comments = [] #: Array[Prism::Comment]
 
+        last_line = node.location.start_line
+
         comments.each do |comment|
           string = comment.slice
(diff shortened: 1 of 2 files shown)


lib/spoom/rbs.rb
```diff
@@ -75,9 +75,19 @@ def node_rbs_comments(node)
...
+        last_line = node.location.start_line
+
         comments.each do |comment|
           string = comment.slice
 
+          comment_line = comment.location.end_line
+
+          if comment_line < last_line - 1
+            break
+          end
```
vinistock, line 87: nit
```suggestion
          break if comment_line < last_line - 1
```

vinistock (approved):

PR #803, opened 2025-10-08
Fix spoom srb assertions translate to preserve trailing comments in correct order

burke:
When translating T.let with trailing comments, the RBS annotation now appears before the comment so Sorbet can properly recognize the type:

 ```ruby
a = T.let(3, Integer) # comment
```

now translates to:

```ruby
a = 3 #: Integer # comment
```

instead of:

```ruby
a = 3 # comment #: Integer
```


burke:
egh I guess I messed up a test, I'll see when I get some time to debug that.

burke:
Oh interesting, it was failing on main too. Well I pushed a fix I think.

PR #808, opened 2025-10-16
Wrap exec command call with shell to avoid corrupted output

Sapphire64:
For one of the issues we've seen internally, wrapping the command call into shell to avoid the issue.

Related: https://github.com/Shopify/team-ruby-dx/issues/1724

### DIFF
--- a/lib/spoom/context/exec.rb
+++ b/lib/spoom/context/exec.rb
@@ -1,6 +1,8 @@
 # typed: strict
 # frozen_string_literal: true
 
+require "shellwords"
+
 module Spoom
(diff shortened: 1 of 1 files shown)


Sapphire64:
@Morriar that was the original version but it adds a complexity of having to handle two types of quotes in the string, so a bit more noisy and error-prone. I can change it to `/bin/sh` if we want to.

Because I was not able to reproduce the issue in local Spoom tests, I personally believe that the fix here is not necessary, and we should find the core issue, while using this branch as a relief point for the team and a guidance on where to look for.

PR #820, opened 2025-11-20
Unset the `DYLD_LIBRARY_PATH` env var when running Sorbet

egiurleo:
There are cases where a dev environment can set the `DYLD_LIBRARY_PATH` env variable in a way that loads a version of the C++ standard library that is not compatible with the system version of the library used by Sorbet. This will cause Sorbet to generate corrupted JSON if the --print=symbol-table-json flag is passed. Unsetting this environment variable forces Sorbet to default to the system version of the C++ standard library, preventing this bug.

**Major question: how to test this?**

I tried testing this by setting an env var on a `Context` object, and then seeing if it would be unset in the context of a `Context.srb` call, but I believe that every invocation of `Context.exec` creates a new shell env that doesn't persist between calls (correct me if I'm wrong here), so none of my attempts were successful. Would appreciate thoughts here!

### DIFF
--- a/lib/spoom/context/sorbet.rb
+++ b/lib/spoom/context/sorbet.rb
@@ -9,10 +9,16 @@ module Spoom
       # Run `bundle exec srb` in this context directory
       #: (*String arg, ?sorbet_bin: String?, ?capture_err: bool) -> ExecResult
       def srb(*arg, sorbet_bin: nil, capture_err: true)
+        # If an environment sets the DYLD_LIBRARY_PATH in a way that depends on a version of the C++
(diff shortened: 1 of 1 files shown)


egiurleo:
@vinistock I think you're right -- they might also have other things in their `ENV` that impact what they want to run here? Issue with this being such an abstract code path that could be used for so many things. I'm going to close this in favor of #808 to avoid this issue.

paracycle:
I think this was the right approach, tbh. I think we can safely remove all env variables from the Sorbet invocation.

PR #824, opened 2025-12-04
Add trusted publishing release workflow

egiurleo:
This workflow is triggered on the push of a new tag, releasing the gem to rubygems.org using trusted publishing and then creating a new GitHub release using the `gh` CLI tool.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,52 @@
+name: Release gem on RubyGems.org and create GitHub release
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+jobs:
+  release_gem:
+    name: Release gem on RubyGems.org
+    if: github.repository == 'Shopify/spoom'
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+      id-token: write
+
+    environment: release
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
+          persist-credentials: false
+
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@8aeb6ff8030dd539317f8e1769a044873b56ea71 # v1.268.0
+        with:
+          bundler-cache: true
+
+      - name: Release gem on RubyGems.org
+        uses: rubygems/release-gem@1c162a739e8b4cb21a676e97b087e8268d8fc40b # v1.1.2
+
+  release_github:
+    name: Create GitHub release
+    if: github.repository == 'Shopify/spoom'
+    needs: release_gem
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
+name: Release gem on RubyGems.org
+
+on:
+  workflow_dispatch:
```
paracycle: Is there a special reason why we are making these `workflow_dispatch` as opposed to being trigged on a special `v*` tag or when a `release` is published like in [here](https://github.com/Shopify/thread_safety/commit/a94e9aebcbb13193302302ba6c23a651ebc742e8)?

I think with a release published, we can even ask another team member/repo member to approve the release publishing, so that we have more security over the process than running a workflow.

Morriar (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
+name: Release gem on RubyGems.org and create GitHub release
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
```
egiurleo: I think I figured out why we were having a hard time releasing when a tag already existed on the RBI gem ([see workflow failure here](https://github.com/Shopify/rbi/actions/runs/19870514915/job/56944729785)) -- [this is the check](https://github.com/ruby/rubygems/blob/master/bundler/lib/bundler/gem_helper.rb#L160) that prevents the `rake release` task from pushing a tag if it already exists. It relies on the `git tag` command, which afaik lists tags that are available locally, so I'm assuming the system can get in a state where a tag exists remotely but not locally, so the workflow will fail when it attempts to push a tag that already exists. The RBI workflow failed correctly.

In general, this should be safe to do -- [see this working correctly on the rdoc repo](https://github.com/ruby/rdoc/actions/runs/20006737341/job/57370113096).

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          bundler-cache: true
+
+      - name: Release gem on RubyGems.org
+        uses: rubygems/release-gem@1c162a739e8b4cb21a676e97b087e8268d8fc40b # v1.1.2
+
+  release_github:
+    name: Create GitHub release
+    if: github.repository == 'Shopify/spoom'
+    needs: release_gem
```
egiurleo: [See the documentation for `needs` here](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-jobs#defining-prerequisite-jobs).

.github/workflows/release.yml
Morriar: Okay so if the tag exists `rake release` doesn't try to push it. Which means the `release-gem` shouldn't fail?

Nice find! Thanks for the explanation.

st0012 (approved):

PR #841, opened 2026-01-08
Migrate all `T::Struct` usages to bare Ruby classes

Morriar:
Also enable `Sorbet/ForbidTStruct` cop after that.

### DIFF
--- a/lib/spoom/sorbet/lsp/structures.rb
+++ b/lib/spoom/sorbet/lsp/structures.rb
@@ -13,11 +13,20 @@ module Spoom
       def accept_printer(printer) = raise NotImplementedError, "Abstract method called"
     end
 
-    class Position < T::Struct
+    class Position
       include PrintableSymbol
 
-      const :line, Integer
-      const :char, Integer
+      #: Integer
+      attr_reader :line
+
+      #: Integer
+      attr_reader :char
+
+      #: (line: Integer, char: Integer) -> void
+      def initialize(line:, char:)
+        @line = line
+        @char = char
+      end
 
       class << self
         #: (Hash[untyped, untyped] json) -> Position
@@ -41,18 +50,27 @@ module Spoom
       end
     end
 
-    class Range < T::Struct
+    class Range
       include PrintableSymbol
 
-      const :start, Position
-      const :end, Position
+      #: Position
+      attr_reader :start_pos
+
+      #: Position
+      attr_reader :end_pos
+
+      #: (start_pos: Position, end_pos: Position) -> void
+      def initialize(start_pos:, end_pos:)
+        @start_pos = start_pos
+        @end_pos = end_pos
+      end
 
       class << self
         #: (Hash[untyped, untyped] json) -> Range
         def from_json(json)
           Range.new(
-            start: Position.from_json(json["start"]),
-            end: Position.from_json(json["end"]),
+            start_pos: Position.from_json(json["start"]),
+            end_pos: Position.from_json(json["end"]),
           )
         end
       end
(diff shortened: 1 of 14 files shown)


amomchilov (approved):
I guess if you migrated to `Data.define` instead, you'd need a whole bunch of RBI files just to specify the sigs, right?

lib/spoom/sorbet/lsp/structures.rb
```diff
@@ -70,41 +88,59 @@ def to_s
...
-    class Location < T::Struct
+    class Location
       include PrintableSymbol
 
-      const :uri, String
-      const :range, LSP::Range
+      #: String
+      attr_reader :uri
+
+      #: Range
```
paracycle: This is quite confusing. Even though it probably resolves to `LSP::Range`, I still find it confusing to see a bare `Range` reference, and I assume it would be a reference to `::Range` and not something else. Should we keep the namespace here?

lib/spoom/sorbet/lsp/structures.rb
```diff
@@ -162,13 +212,28 @@ def to_s
...
 
-    class Diagnostic < T::Struct
+    class Diagnostic
       include PrintableSymbol
 
-      const :range, LSP::Range
-      const :code, Integer
-      const :message, String
-      const :information, Object
+      #: Range
```
paracycle: Same comment as above

lib/spoom/sorbet/lsp/structures.rb
```diff
@@ -221,7 +314,7 @@ def from_json(json)
       # @override
       #: (SymbolPrinter printer) -> void
       def accept_printer(printer)
-        h = serialize.hash
+        h = hash
```
paracycle: Interesting! I am noticing this code now and I am not sure if this will always work the way you intend it to. The return value of `hash` isn't necessarily unique, so you can't rely on it (solely) for Set inclusion.
> `a.eql?(b)` implies `a.hash == b.hash`

_from https://docs.ruby-lang.org/en/3.4/Object.html#method-i-hash_

but the converse isn't necessarily true, i.e. `a.hash == b.hash` does not necessarily imply `a.eql?(b)`.

paracycle (approved):
I have a few minor comments, but looks good to me

lib/spoom/sorbet/lsp/structures.rb
Morriar: Addressed in dbd3f47d by restoring the explicit LSP::Range type.

lib/spoom/sorbet/lsp/structures.rb
Morriar: Addressed in 500c9de9. SymbolPrinter now stores a deep structural key in the Set instead of an integer hash, preserving value-based deduplication while allowing distinct values with colliding hashes. Added regression coverage for both cases.

lib/spoom/sorbet/lsp/structures.rb
Morriar: Addressed in dbd3f47d by restoring the explicit LSP::Range type.

lib/spoom/sorbet/lsp/structures.rb
```diff
@@ -194,15 +259,43 @@ def to_s
...
+      #: String?
+      attr_reader :detail
+
+      #: Integer
+      attr_reader :kind
+
+      #: Location?
+      attr_reader :location
+
+      #: Range?
```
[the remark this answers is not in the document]
Morriar: Addressed in dbd3f47d by explicitly using LSP::Range here as well.

PR #850, opened 2026-01-13
Fix test runner and output

paracycle:
Now that this gem is on Minitest 6, we are able to use its own runner for running tests. So, I created a binstub for the `minitest` command exported from the `minitest` gem and renamed it to `bin/test`. This way, we can continue to use the (almost) same command line interface for running tests, but we also get running tests by line number too 🎉

Additionally, since Minitest no longer automatically registers plugins, we need to fix up how we register Minitest reporters that we were registering.


test/test_helper.rb
```diff
@@ -50,5 +50,7 @@ def default_spoom_test_gemfile
 
 require "minitest/autorun"
 require "minitest/reporters"
+require "minitest/minitest_reporter_plugin"
+Minitest.register_plugin(:minitest_reporter)
```
Morriar: Do we still need the reporter?

paracycle: Yes! This is for registering Minitest Reporters' plugin to Minitest (which used to be automatic previously), but we still need to set the reporter, etc after that.

In the next release of `minitest-reporters` this should be done automatically inside the library. For now, we need to do it explicitly.

paracycle: I haven't seen this one be that buggy, tbh, we can always change it if it is.

PR #861, opened 2026-02-17
Fix RBS rewriting to preserve line numbers

### LINKED ISSUE #787: RBS Rewriting doesn't preserve line numbers
**Describe the bug**

When using `Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sig`, the line numbers aren't preserved. If the rewritten code is used in lieu of the original source, errors and stacktraces don't line up with the original source.

**To Reproduce**

```ruby
print Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs(<<~RB, file: "test.rb")
  # typed: true
 
  #: (
  #| Integer
  #| ) -> void
  def foo(i); end
RB
```

```
# typed: true

sig { params(i: Integer).void }
def foo(i); end
```

`foo` is defined at line 4 instead of line 6.

**Expected behavior**

Ideally 

```
# typed: true

sig { params(
  i: Integer
  ).void }
def foo(i); end
```

But this is fine too

```
# typed: true

sig { params(i: Integer).void }


def foo(i); end
```

Morriar:
Fixes #787 

## Problem

When using `Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sig`, line numbers weren't preserved. Multi-line RBS comments were replaced with single-line sigs, causing method definitions to shift up in the file. This made errors and stacktraces misalign with the original source.

**Example:**
```ruby
# Original (method at line 6)
#: (
#| Integer
#| ) -> void
def foo(i); end

# After translation (method moved to line 4)
sig { params(i: Integer).void }
def foo(i); end
```

## Solution

Added a `preserve_line_numbers` helper that:
1. Calculates how many lines the original RBS comment spans
2. Calculates how many lines the generated sig will have
3. Adds blank lines after the sig to match the original line count

**Result:**
```ruby
# After fix (method stays at line 6)
sig { params(i: Integer).void }


def foo(i); end
```

## Testing

* All existing tests pass
* Updated `test_translate_to_rbi_multiline_sigs` to expect preserved line numbers
* Manually tested with various RBS comment sizes (single-line, multi-line, complex signatures)

## Changes

* `lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb`:
  * Added `preserve_line_numbers` helper method
  * Updated `rewrite_def` to use line preservation
  * Updated `visit_attr` to use line preservation
* `test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb`:
  * Updated `test_translate_to_rbi_multiline_sigs` expectations

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
@@ -73,6 +73,21 @@ module Spoom
 
         private
 
+        # Preserves line numbers by adding blank lines after the replacement string
+        # to match the number of lines in the original location
+        #: (Prism::Location, String) -> String
+        def preserve_line_numbers(location, replacement_string)
+          original_lines = location.end_line - location.start_line + 1
+          replacement_lines = replacement_string.lines.count
+
+          if original_lines > replacement_lines
+            padding = "\n" * (original_lines - replacement_lines)
+            replacement_string + padding
+          else
+            replacement_string
+          end
+        end
+
         #: (Prism::CallNode) -> void
         def visit_attr(node)
           comments = node_rbs_comments(node)
@@ -100,10 +115,13 @@ module Spoom
 
             apply_member_annotations(comments.method_annotations, sig)
 
+            sig_string = sig.string(max_line_length: @max_line_length)
+            replacement = preserve_line_numbers(signature.location, sig_string)
+
             @rewriter << Source::Replace.new(
               signature.location.start_offset,
               signature.location.end_offset,
-              sig.string(max_line_length: @max_line_length),
+              replacement,
             )
           rescue ::RBS::ParsingError, ::RBI::Error
             # Ignore signatures with errors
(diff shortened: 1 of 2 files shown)


lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -73,6 +73,21 @@ def visit_call_node(node)
...
+        #: (Prism::Location, String) -> String
+        def preserve_line_numbers(location, replacement_string)
+          original_lines = location.end_line - location.start_line + 1
+          replacement_lines = replacement_string.lines.count
+
+          if original_lines > replacement_lines
+            padding = "\n" * (original_lines - replacement_lines)
+            replacement_string + padding
+          else
+            replacement_string
```
paracycle: What happens if `original_lines < replacement_lines`? Could that ever happen? Should we handle that explicitly and `raise` if that is not expected?

amomchilov:
Superseded by https://github.com/Shopify/spoom/pull/940

PR #879, opened 2026-03-30
Add --ignore-errors option to `spoom srb tc`

Morriar:
## Summary

* Add `--ignore-errors` option to `spoom srb tc` to ignore known errors by matching `error_code:file:line` from an ignore file
* Default ignore file location: `sorbet/ignored_errors.cfg` (auto-detected when present)
* Ignored errors are hidden from output, shown as a count in the summary (e.g. `Errors: 2 ignored, 7 total`), and don't cause a non-zero exit code
* Supports `#` comments and blank lines in the ignore file

## Ignore file format

```
# Comments supported
5002:errors/errors.rb:5
7003:errors/errors.rb:11
```

## Test plan

* [x] All errors ignored → exit 0, summary shows `X ignored, Y total`
* [x] Partial match → exit 1, only non-ignored errors displayed
* [x] Custom path via `--ignore-errors=path`
* [x] Comments and blank lines in ignore file
* [x] No ignore file → existing behavior unchanged (29 tests pass)

### DIFF
--- a/lib/spoom/cli/srb/tc.rb
+++ b/lib/spoom/cli/srb/tc.rb
@@ -25,6 +25,7 @@ module Spoom
         option :junit_output_path, type: :string, desc: "Output failures to XML file formatted for JUnit"
         option :sorbet, type: :string, desc: "Path to custom Sorbet bin"
         option :sorbet_options, type: :string, default: "", desc: "Pass options to Sorbet"
+        option :ignore_errors, type: :string, desc: "Path to ignored errors file (default: sorbet/ignored_errors.cfg)"
         def tc(*paths_to_select)
           context = context_requiring_sorbet!
           limit = options[:limit]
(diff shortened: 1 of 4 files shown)


lib/spoom/cli/srb/tc.rb
```diff
@@ -84,6 +94,32 @@ def tc(*paths_to_select)
...
+                true
+              end
+            end
+            ignored_count = ignored.size
+            errors = active
+
+            stale = ignored_errors - matched
+            unless stale.empty?
+              stale.each do |entry|
+                say_error("Stale entry in ignore file: #{entry[0]}:#{entry[1]}:#{entry[2]}", status: nil)
```
KaanOzkan, line 117: Looks like we exit early `if result.status` above. So stale information is only printed if there are errors, is this intentional? I feel like we should always try to keep the list up to date.

PR #901, opened 2026-04-14
Preserve inline comments inside sig blocks during RBS translation

### LINKED ISSUE #900: `srb sigs translate` removes comments during translation
`spoom srb sigs translate` removes some comments during translation.

```rb
sig do
  params(
    # The prefix to the redis key
    prefix: String,
    # The TTL set in Redis by default, unless overridden in the session.options[:expire_after]
    ttl: String
  ).void
end
```

KaanOzkan:
Comments inside `sig do params(...) end` blocks were silently dropped because the entire sig node byte range was replaced with the RBS string. Collect any comments within the sig node's range and prepend them to the output before the `#:` annotation.

Resolves https://github.com/Shopify/spoom/issues/900

### DIFF
--- a/test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
+++ b/test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
@@ -466,6 +466,59 @@ module Spoom
           RBS
         end
 
+        def test_translate_to_rbs_sig_with_inline_param_comments
+          contents = <<~RB
+            sig do
+              params(
+                # First param
+                a: Integer,
+                # Second param
+                b: String
+              ).void
+            end
+            def foo(a, b); end
+
+            sig do
+              # A comment
+              returns(Integer)
+            end
+            attr_reader :x
+          RB
+
+          assert_equal(<<~RBS, sorbet_sigs_to_rbs_comments(contents))
+            # First param
+            # Second param
+            #: (Integer a, String b) -> void
+            def foo(a, b); end
+
+            # A comment
+            #: Integer
+            attr_reader :x
+          RBS
+        end
+
+        def test_translate_to_rbs_sig_with_inline_param_comments_indented
+          contents = <<~RB
+            class Foo
+              sig do
+                params(
+                  # First param
+                  a: Integer
+                ).void
+              end
+              def foo(a); end
+            end
+          RB
+
+          assert_equal(<<~RBS, sorbet_sigs_to_rbs_comments(contents))
+            class Foo
+              # First param
+              #: (Integer a) -> void
+              def foo(a); end
+            end
+          RBS
+        end
+
         def test_translate_to_rbs_defs_within_send
           contents = <<~RB
             sig { void }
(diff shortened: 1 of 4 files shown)


test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
```diff
@@ -466,6 +466,37 @@ def foo(param1:, param2:); end
...
+
+          assert_equal(<<~RBS, sorbet_sigs_to_rbs_comments(contents))
+            # First param
+            # Second param
+            #: (Integer a, String b) -> void
+            def foo(a, b); end
+
+            # A comment
+            #: Integer
+            attr_reader :x
```
Morriar, line 496: Could we add a test with a sig inside a class (indented)? The logic uses `sig_node.location.start_column` to build the indent prefix between comments, and I want to make sure that's correct when the sig isn't at column 0.

Something like:

```ruby
def test_translate_to_rbs_sig_with_inline_param_comments_indented
  contents = <<~RB
    class Foo
      sig do
        params(
          # First param
          a: Integer
        ).void
      end
      def foo(a); end
    end
  RB

  assert_equal(<<~RBS, sorbet_sigs_to_rbs_comments(contents))
    class Foo
      # First param
      #: (Integer a) -> void
      def foo(a); end
    end
  RBS
end
```

test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
```diff
@@ -466,6 +466,37 @@ def foo(param1:, param2:); end
...
+              # A comment
+              returns(Integer)
+            end
+            attr_reader :x
+          RB
+
+          assert_equal(<<~RBS, sorbet_sigs_to_rbs_comments(contents))
+            # First param
+            # Second param
+            #: (Integer a, String b) -> void
```
Morriar, line 491: I think we should force multi line rendering when we find comments. 

Consider this:

```rb
params(
  # Some documentation
  a: Integer
).void
def foo(a); end
```

when translated we get this:

```rb
# Some documentation
#: (a) -> void
def foo(a); end
```

it will look as if the documentation is about `foo` and not `a`.

test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
KaanOzkan: Done.

test/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments_test.rb
KaanOzkan: Hmm yeah this makes sense. I think we can either:
1. Associate comments with parameters in `RBI` and teach `RBSPrinter` about outputting it
2. Keep doing the string post-processing in Spoom, similar to the new `add_inline_sig_comments`. We could figure out the multiline aspect by looking at the indentation and insert the comment to the relevant line. We'd also need to add a new flag or set `max_line_length` to 0 for the `RBSPrinter` to print multiline comment.

I'm leaning towards 1 since we're getting too much into the printing territory. Wdyt?

Morriar: Yes, 1. 👍

PR #917, opened 2026-05-12
Skip RBS rewriter when file does not contain RBS syntax

### LINKED ISSUE #916: Skip RBS rewriter when a file doesn't contain any RBS syntax
Before we waste the time parsing with Prism, running our RBS rewriting visitor, etc., we can do a quick scan through the file to see if it contains any RBS syntax at all. If not, we can just return it as it was.

This was already implemented into Tapioca, but would better live here in Spoom directly: https://github.com/Shopify/tapioca/pull/2616

dejmedus:
- closes https://github.com/Shopify/spoom/issues/916
- moved from tapioca https://github.com/Shopify/tapioca/pull/2616

If a file does not contain typed file markers (ex `# typed: true`) or RBS syntax we can skip attempting to translate it. If a file has not been rewritten we can exclude it from the count of translated files

### DIFF
--- a/lib/spoom/cli/srb/sigs.rb
+++ b/lib/spoom/cli/srb/sigs.rb
@@ -213,8 +213,10 @@ module Spoom
                 contents = contents.force_encoding(encoding)
               end
 
-              contents = block.call(file, contents)
-              File.write(file, contents)
+              new_contents = block.call(file, contents)
+              next if new_contents == contents
+
+              File.write(file, new_contents)
               transformed_count += 1
             rescue RBI::Error => error
               say_warning("Can't parse #{file}: #{error.message}")
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
@@ -7,6 +7,33 @@ module Spoom
       class RBSCommentsToSorbetSigs < Translator
         include Spoom::RBS::ExtractRBSComments
 
+        RBS_ANNOTATION_MARKERS = [
+          "# @abstract",
+          "# @interface",
+          "# @sealed",
+          "# @final",
+          "# @requires_ancestor:",
+          "# @override",
+          "# @overridable",
+          "# @without_runtime",
+        ].freeze #: Array[String]
+        RBS_REWRITE_PATTERN = Regexp.union(["#:", "#|", *RBS_ANNOTATION_MARKERS]).freeze #: Regexp
+        private_constant :RBS_ANNOTATION_MARKERS, :RBS_REWRITE_PATTERN
+
+        class << self
+          #: (String source) -> bool
+          def contains_rbs_syntax?(source)
+            Sigils.contains_valid_sigil?(source) && source.match?(RBS_REWRITE_PATTERN)
+          end
+
+          #: (String ruby_contents, file: String, ?max_line_length: Integer?) -> String
+          def rewrite_if_needed(ruby_contents, file:, max_line_length: nil)
+            return ruby_contents unless contains_rbs_syntax?(ruby_contents)
+
+            new(ruby_contents, file:, max_line_length:).rewrite
+          end
+        end
+
         #: (String, file: String, ?max_line_length: Integer?) -> void
         def initialize(ruby_contents, file:, max_line_length: nil)
           super(ruby_contents, file: file)
--- a/rbi/spoom.rbi
+++ b/rbi/spoom.rbi
@@ -2823,6 +2823,9 @@ module Spoom::Sorbet::Sigils
     sig { params(path_list: T::Array[::String], new_strictness: ::String).returns(T::Array[::String]) }
     def change_sigil_in_files(path_list, new_strictness); end
 
+    sig { params(content: ::String).returns(T::Boolean) }
+    def contains_valid_sigil?(content); end
+
     sig { params(path: T.any(::Pathname, ::String)).returns(T.nilable(::String)) }
     def file_strictness(path); end
 
@@ -2938,7 +2941,18 @@ class Spoom::Sorbet::Translate::RBSCommentsToSorbetSigs < ::Spoom::Sorbet::Trans
 
   sig { params(node: ::Prism::CallNode).void }
   def visit_attr(node); end
+
+  class << self
+    sig { params(source: ::String).returns(T::Boolean) }
+    def contains_rbs_syntax?(source); end
+
+    sig { params(ruby_contents: ::String, file: ::String, max_line_length: T.nilable(::Integer)).returns(::String) }
+    def rewrite_if_needed(ruby_contents, file:, max_line_length: T.unsafe(nil)); end
+  end
 end
+
+Spoom::Sorbet::Translate::RBSCommentsToSorbetSigs::RBS_ANNOTATION_MARKERS = T.let(T.unsafe(nil), Array)
+Spoom::Sorbet::Translate::RBSCommentsToSorbetSigs::RBS_REWRITE_PATTERN = T.let(T.unsafe(nil), Regexp)
 
 class Spoom::Sorbet::Translate::SorbetAssertionsToRBSComments < ::Spoom::Sorbet::Translate::Translator
   sig do
--- a/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
+++ b/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
@@ -678,6 +678,141 @@ module Spoom
           RB
         end
 
+        def test_contains_rbs_syntax_returns_true_for_supported_rbs_annotations
+          [
+            "# @abstract",
+            "# @interface",
+            "# @sealed",
+            "# @final",
+            "# @requires_ancestor:",
+            "# @override",
+            "# @override(allow_incompatible: true)",
+            "# @override(allow_incompatible: false)",
+            "# @override(allow_incompatible: :visibility)",
+            "# @overridable",
+            "# @without_runtime",
+          ].each do |marker|
+            assert(
+              RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB),
+                # typed: true
+
+                #{marker}
+                class Foo; end
+              RB
+              "#contains_rbs_syntax? should return true for files containing #{marker}",
+            )
+          end
+        end
+
+        def test_contains_rbs_syntax_returns_true_for_supported_typed_sigils
+          [
+            "# typed: ignore",
+            "# typed: false",
+            "# typed: true",
+            "# typed: strict",
+            "# typed: strong",
+            "# typed: __STDLIB_INTERNAL",
+          ].each do |sigil|
+            assert(
+              RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB),
+                #{sigil}
+
+                #: -> void
+                def foo; end
+              RB
+              "#contains_rbs_syntax? should return true for files containing #{sigil}",
+            )
+          end
+        end
+
+        def test_contains_rbs_syntax_returns_true_when_typed_sigil_is_after_other_magic_comments
+          assert(RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB))
+            # frozen_string_literal: true
+            # typed: true
+
+            class Foo
+              #: -> String
+              def foo; end
+            end
+          RB
+
+          assert(RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB))
+            # frozen_string_literal: true
+
+            # typed: true
+
+            class Foo
+              #: -> String
+              def foo; end
+            end
+          RB
+        end
+
+        def test_contains_rbs_syntax_returns_true_for_rbs_comments
+          assert(RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB))
+            # typed: true
+
+            class Foo
+              #: -> String
+              def foo; end
+            end
+          RB
+        end
+
+        def test_contains_rbs_syntax_returns_true_for_multiline_rbs_comments
+          assert(RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB))
+            # typed: true
+
+            class Foo
+              #: -> Array[
+              #| String
+              #| ]
+              def foo; end
+            end
+          RB
+        end
+
+        def test_contains_rbs_syntax_returns_false_for_files_without_typed_sigil
+          refute(RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB))
+            #: -> void
+            def foo; end
+          RB
+        end
+
+        def test_contains_rbs_syntax_returns_false_for_files_without_rbs_syntax
+          refute(RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB))
+            # typed: true
+
+            class Foo
+              def foo; end
+            end
+          RB
+        end
+
+        def test_contains_rbs_syntax_returns_false_for_unrelated_yard_tags
+          refute(RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB))
+            # typed: true
+
+            # @param value [String]
+            # @return [String]
+            def foo(value); end
+          RB
+        end
+
+        def test_rewrite_does_not_call_new_for_files_without_rbs_syntax
+          source = <<~RB
+            # typed: true
+
+            class Foo
+              def foo; end
+            end
+          RB
+
+          RBSCommentsToSorbetSigs.stub(:new, ->(*) { flunk("should not be called") }) do
+            assert_equal(source, RBSCommentsToSorbetSigs.rewrite_if_needed(source, file: "test.rb"))
+          end
+        end
+
         private
 
         #: (String, ?max_line_length: Integer?) -> String
--- a/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
+++ b/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
@@ -817,7 +817,7 @@ module Spoom
 
         #: (String, ?max_line_length: Integer?) -> String
         def rbs_comments_to_sorbet_sigs(ruby_contents, max_line_length: nil)
-          Translate.rbs_comments_to_sorbet_sigs(ruby_contents, file: "test.rb", max_line_length: max_line_length)
+          RBSCommentsToSorbetSigs.new(ruby_contents, file: "test.rb", max_line_length: max_line_length).rewrite
         end
       end
     end
--- a/lib/spoom/sorbet/sigils.rb
+++ b/lib/spoom/sorbet/sigils.rb
@@ -44,6 +44,13 @@ module Spoom
           SIGIL_REGEXP.match(content)&.[](1)
         end
 
+        # returns true if the passed content contains a valid sigil
+        #: (String content) -> bool
+        def contains_valid_sigil?(content)
+          strictness = strictness_in_content(content)
+          !!strictness && valid_strictness?(strictness)
+        end
+
         # returns a string which is the passed content but with the sigil updated to a new strictness
         #: (String content, String new_strictness) -> String
         def update_sigil(content, new_strictness)
--- a/lib/spoom/sorbet/translate.rb
+++ b/lib/spoom/sorbet/translate.rb
@@ -55,7 +55,7 @@ module Spoom
         # It also handles type members and class annotations.
         #: (String ruby_contents, file: String, ?max_line_length: Integer?) -> String
         def rbs_comments_to_sorbet_sigs(ruby_contents, file:, max_line_length: nil)
-          RBSCommentsToSorbetSigs.new(ruby_contents, file: file, max_line_length: max_line_length).rewrite
+          RBSCommentsToSorbetSigs.rewrite_if_needed(ruby_contents, file: file, max_line_length: max_line_length)
         end
 
         # Converts all `T.let` and `T.cast` nodes to RBS comments in the given Ruby code.
--- a/test/spoom/cli/srb/sigs_test.rb
+++ b/test/spoom/cli/srb/sigs_test.rb
@@ -351,6 +351,8 @@ module Spoom
 
         def test_translate_from_rbs_to_rbi
           @project.write!("file.rb", <<~RB)
+            # typed: true
+
             #: -> void
             def foo; end
           RB
@@ -366,6 +368,8 @@ module Spoom
           assert(result.status)
 
           assert_equal(<<~RB, @project.read("file.rb"))
+            # typed: true
+
             sig { void }
             def foo; end
           RB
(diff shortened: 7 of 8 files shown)


test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -678,6 +784,162 @@ class Range
           RB
         end
 
+        def test_should_rewrite_returns_true_for_supported_typed_sigils
```
amomchilov: Two more cases to test:

1. Don't trigger if `# typed: true` exists later in the file. This is unlikely, but it's a performance improvement to ensure we're not scanning through the whole file
2. Trigger if there's multiple magic comments, but `typed` isn't first:

    ```ruby
    # frozen_string_literal: true
    # typed: true
    ```

dejmedus: ~~> Don't trigger if # typed: true exists later in the file~~

~~Good call! I updated to only check for the typed marker in the magic comments block~~

Refactored to instead use `strictness_in_content` and `valid_strictness` checks from [`/sorbet/sigils.rb`](https://github.com/Shopify/spoom/pull/917/changes#diff-5ab95c38bcfa1d2a2f11410db1772f3d8d3853c70e2045e19f0538496de267bfR49)

lib/spoom/cli/srb/sigs.rb
```diff
@@ -213,8 +213,10 @@ def transform_files(files, &block)
                 contents = contents.force_encoding(encoding)
               end
 
-              contents = block.call(file, contents)
-              File.write(file, contents)
+              new_contents = block.call(file, contents)
+              next if new_contents == contents
+
+              File.write(file, new_contents)
               transformed_count += 1
```
dejmedus, line 220: This  will only count changed files in the `Translated signatures in <x> files.` message, but I'm happy to drop the commit if we want to leave as is

paracycle (commented):
Shouldn't all this logic be encapsulated inside the `RBSCommentsToSorbetSigs` class, which already stores the `ruby_contents` as an ivar, and wouldn't need to pass it to a `should_rewrite?` method. The downside is just an extra object allocation only to return the same contents, but I think that's fine.

lib/spoom/cli/srb/sigs.rb
amomchilov: I like it.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -7,6 +7,27 @@ module Translate
...
+          "# @overridable",
+          "# @without_runtime",
+        ].freeze #: Array[String]
+        RBS_REWRITE_PATTERN = Regexp.union(["#:", "#|"] + RBS_ANNOTATION_MARKERS).freeze #: Regexp
+
+        class << self
+          #: (String source) -> bool
+          def should_rewrite?(source)
+            source.match?(TYPED_FILE_PATTERN) && source.match?(RBS_REWRITE_PATTERN)
+          end
```
amomchilov, line 34: To what Ufuk was saying, maybe we can add a `rewrite` class method:


```suggestion
          end
          
          #: (String ruby_contents, file: String, ?max_line_length: Integer?) -> String
          def rewrite(ruby_contents, file:, max_line_length: nil)
            # fast path, skip parsing/visiting if there's no RBS syntax at all
            return ruby_contents if should_rewrite?(ruby_contents)
            
            new(ruby_contents, file:, max_line_length:).rewrite
          end
```

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -7,6 +7,27 @@ module Translate
...
+          "# @abstract",
+          "# @interface",
+          "# @sealed",
+          "# @final",
+          "# @requires_ancestor:",
+          "# @override",
+          "# @overridable",
+          "# @without_runtime",
+        ].freeze #: Array[String]
+        RBS_REWRITE_PATTERN = Regexp.union(["#:", "#|"] + RBS_ANNOTATION_MARKERS).freeze #: Regexp
```
amomchilov: tiny opt:

```suggestion
        RBS_REWRITE_PATTERN = Regexp.union(["#:", "#|", *RBS_ANNOTATION_MARKERS]).freeze #: Regexp
```

Spares one allocation :)

```ruby
require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "benchmark-memory"
end

a = [3, 4, 5]
Benchmark.memory do |x|
  x.report("    +") { [1, 2] + a }
  x.report("splat") { [1, 2, *a] }
  x.compare!
end

# Calculating -------------
#      +   120.000  memsize
#            2.000  objects
#  splat   208.000  memsize
#            1.000  objects
```

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -7,6 +7,27 @@ module Translate
...
+
+        class << self
+          #: (String source) -> bool
+          def should_rewrite?(source)
```
amomchilov: if this is going to be public API, perhaps `contains_rbs_syntax?` would be more clear to callers

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -678,6 +784,197 @@ class Range
...
+              def foo; end
+            end
+          RB
+
+          RBSCommentsToSorbetSigs.stub(:new, ->(*) { flunk("should not be called") }) do
+            assert_equal(source, rbs_comments_to_sorbet_sigs(source))
+          end
+        end
+
+        def test_calls_new_for_files_with_rbs_syntax
```
amomchilov: eh, I don't think this is a useful test. it's implied from the other more end-to-endy tests.

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -678,6 +784,197 @@ class Range
...
+            "# @without_runtime",
+          ].each do |marker|
+            assert(
+              RBSCommentsToSorbetSigs.should_rewrite?(<<~RB),
+                # typed: true
+
+                #{marker}
+                class Foo; end
+              RB
+              marker,
```
amomchilov:
```suggestion
              "#should_rewrite? should return true for files containing #{marker}",
```

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -678,6 +784,197 @@ class Range
...
+        end
+
+        def test_should_rewrite_returns_true_for_supported_rbs_annotations
+          [
+            "# @abstract",
+            "# @interface",
+            "# @sealed",
+            "# @final",
+            "# @requires_ancestor:",
+            "# @override",
```
amomchilov: I think you can just move these here and remove the next test

```suggestion
            "# @override",
            "# @override(allow_incompatible: true)",
            "# @override(allow_incompatible: false)",
            "# @override(allow_incompatible: :visibility)",
```

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -678,6 +784,197 @@ class Range
...
+
+                #{marker}
+                def foo; end
+              RB
+              marker,
+            )
+          end
+        end
+
+        def test_should_rewrite_returns_false_for_files_without_rbs_syntax
```
amomchilov: Can you please reorder these, so related tests are better grouped together?

E.g. the `def test_should_rewrite_returns_false_when_the_file_is_not_typed` should be in this area, where the other `test_should_rewrite_returns_false_for_` tests are

rbi/spoom.rbi
```diff
@@ -2938,8 +2938,17 @@ class Spoom::Sorbet::Translate::RBSCommentsToSorbetSigs < ::Spoom::Sorbet::Trans
...
   sig { params(node: ::Prism::CallNode).void }
   def visit_attr(node); end
+
+  class << self
+    sig { params(source: ::String).returns(T::Boolean) }
+    def should_rewrite?(source); end
+  end
 end
 
+Spoom::Sorbet::Translate::RBSCommentsToSorbetSigs::RBS_ANNOTATION_MARKERS = T.let(T.unsafe(nil), Array)
```
amomchilov, line 2954: Let's mark these with `private_constant` where they're defined, so we don't expose them to the public.

amomchilov (commented):
Few small points, but it's shaping up!

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
dejmedus: Makes sense, thank you both!

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
dejmedus: Nice! 💯

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
dejmedus: Moved things around:)

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -678,6 +784,141 @@ class Range
...
+
+                #: -> void
+                def foo; end
+              RB
+              "#contains_rbs_syntax? should return true for files containing #{sigil}",
+            )
+          end
+        end
+
+        def test_contains_rbs_syntax_returns_true_when_typed_sigils_follow_magic_comments
```
amomchilov:
```suggestion
        def test_contains_rbs_syntax_returns_true_when_typed_sigil_is_after_other_magic_comments
```

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -678,6 +784,141 @@ class Range
...
+          refute(RBSCommentsToSorbetSigs.contains_rbs_syntax?(<<~RB))
+            # typed: true
+
+            class Foo
+              def foo; end
+            end
+          RB
+        end
+
+        def test_contains_rbs_syntax_returns_false_for_unrelated_yard_tags
```
amomchilov, line 792: Good call

amomchilov (approved):
LGTM pending the unresolved comments

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -7,6 +7,33 @@ module Translate
...
+        private_constant :RBS_ANNOTATION_MARKERS, :RBS_REWRITE_PATTERN
+
+        class << self
+          #: (String source) -> bool
+          def contains_rbs_syntax?(source)
+            Sigils.contains_valid_sigil?(source) && source.match?(RBS_REWRITE_PATTERN)
+          end
+
+          #: (String ruby_contents, file: String, ?max_line_length: Integer?) -> String
+          def rewrite(ruby_contents, file:, max_line_length: nil)
```
Morriar:
```suggestion
          def rewrite_if_needed(ruby_contents, file:, max_line_length: nil)
```

or something like this

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -44,6 +44,8 @@ def foo(a, b)
 
         def test_translate_to_rbi_method_sigs
           contents = <<~RB
+            # typed: true
```
Morriar: Can we instead call the underlying `new.rewrite` instead so we don't have to rewrite all the tests?

dejmedus: ~~Can do:) Updated (I just left the few in the cli `sigs_test.rb` file)~~

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -7,6 +7,33 @@ module Translate
...
+          end
+
+          #: (String ruby_contents, file: String, ?max_line_length: Integer?) -> String
+          def rewrite_if_needed(ruby_contents, file:, max_line_length: nil)
+            return ruby_contents unless contains_rbs_syntax?(ruby_contents)
+
+            new(ruby_contents, file:, max_line_length:).rewrite
+          end
+        end
+
```
paracycle, line 36: I am not opinionated about this but I would have prefered:

```suggestion
        #: (String source) -> bool
        def contains_rbs_syntax?
          Sigils.contains_valid_sigil?(@ruby_contents) && @ruby_contents.match?(RBS_REWRITE_PATTERN)
        end

        # @override
        #: () -> String
        def rewrite
          return @ruby_contents unless contains_rbs_syntax?

          super
        end
```

paracycle: In summary, I think we are unnecessarily pulling logic into class methods and passing values around when the value we are interested in is being passed into the constructor of our class, and no-one cares what exactly `rewrite` does internally (i.e. in this case it decides to return the original buffer intact).

paracycle: I notice that we eagerly do parsing in the `initialize` of `Translator`, but we could also short-circuit the call to `super` in this class's `initialize` method by doing the sigil and rewrite checks in the initializer.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -7,6 +7,33 @@ module Translate
...
         include Spoom::RBS::ExtractRBSComments
 
+        RBS_ANNOTATION_MARKERS = [
+          "# @abstract",
+          "# @interface",
+          "# @sealed",
+          "# @final",
+          "# @requires_ancestor:",
+          "# @override",
+          "# @overridable",
```
amomchilov, line 17: Random thought I had: does factoring out the common prefixes produce a more faster regular expression? E.g. `overrid(?:e|able)` instead of checking for the two whole words separately.

In short, yes, but barely. Onigmo doesn't optimize out common prefixes in the NFA it creates, but it doesn't make a real world difference here.
 
Got Claude to benchmark it for me:

```ruby
require "benchmark"

# loading corpus... 101996 files, 616MB, loaded in 15.09s
# sanity-checking match equivalence... 19104 total matches across corpus
#
#                     # user     system      total        real
# flat              0.536375   0.006953   0.543328 (  0.552176)
# factored          0.530740   0.005332   0.536072 (  0.544158)
#
# factored is 1.01x faster than flat (real time)

MARKERS = [
  "# @abstract",
  "# @interface",
  "# @sealed",
  "# @final",
  "# @requires_ancestor:",
  "# @override",
  "# @overridable",
  "# @without_runtime",
].freeze


FLAT     = Regexp.union(*MARKERS)
FACTORED = /\# @(?:abstract|interface|sealed|final|requires_ancestor:|overrid(?:e|able)|without_runtime)/

raise "regexes diverge on hits"   unless MARKERS.all? { |m| FLAT =~ m && FACTORED =~ m }
raise "regexes diverge on misses" if FLAT =~ "# @other" || FACTORED =~ "# @other"

CORPUS_GLOB = "/your/test/corpus/**/*.rb"

print "loading corpus... "
load_start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
CORPUS = Dir.glob(CORPUS_GLOB).map { |path| File.read(path) rescue nil }.compact.freeze
load_elapsed = Process.clock_gettime(Process::CLOCK_MONOTONIC) - load_start
total_bytes = CORPUS.sum(&:bytesize)
puts "#{CORPUS.size} files, #{total_bytes / 1_000_000}MB, loaded in #{load_elapsed.round(2)}s"

print "sanity-checking match equivalence... "
flat_hits     = CORPUS.sum { |s| s.scan(FLAT).size }
factored_hits = CORPUS.sum { |s| s.scan(FACTORED).size }
raise "regexes diverge: flat=#{flat_hits} factored=#{factored_hits}" unless flat_hits == factored_hits
puts "#{flat_hits} total matches across corpus"

flat_time     = Benchmark.measure("flat")     { CORPUS.each { |s| s.scan(FLAT) } }
factored_time = Benchmark.measure("factored") { CORPUS.each { |s| s.scan(FACTORED) } }

slower, faster = [flat_time, factored_time].sort_by(&:real).reverse
ratio = slower.real / faster.real
faster_name = faster.equal?(flat_time) ? "flat" : "factored"
slower_name = slower.equal?(flat_time) ? "flat" : "factored"

puts <<~RESULTS

                  #{Benchmark::CAPTION}
  flat            #{flat_time.to_s.strip}
  factored        #{factored_time.to_s.strip}

  #{faster_name} is #{ratio.round(2)}x faster than #{slower_name} (real time)
RESULTS
```

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -7,13 +7,42 @@ module Translate
...
+        end
+
+        # @override
+        #: () -> String
+        def rewrite
+          return @ruby_contents unless contains_rbs_syntax?
+
+          super
+        end
+
```
paracycle:
```suggestion
        def initialize(ruby_contents, file:, max_line_length: nil)
          @ruby_contents = ruby_contents

          if contains_rbs_syntax?
            super(ruby_contents, file: file)
            
            @max_line_length = max_line_length
          else
            @no_rewrite = true
          end
        end

        #: () -> bool
        def contains_rbs_syntax?
          Sigils.contains_valid_sigil?(@ruby_contents) && @ruby_contents.match?(RBS_REWRITE_PATTERN)
        end

        # @override
        #: () -> String
        def rewrite
          return @ruby_contents if @no_rewrite

          super
        end

```

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
dejmedus: I think I was thinking that I needed to exit early before we initialized to prevent hitting `parse_ruby` and that `rewrite` needed that path to have run already, but I think this is clicking for me now, thanks very much for the examples!

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
dejmedus: Thank you, [updated](https://github.com/Shopify/spoom/pull/917/changes/0b8d325b981b243e2d64991e9a0ac98945408436#diff-3a035aa7b7631b6ac64e7316797385dd56355646b95ce2018576021bd72ff712R24)!

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
dejmedus: Sorry scratch that, [the newest commit](https://github.com/Shopify/spoom/pull/917/commits/0b8d325b981b243e2d64991e9a0ac98945408436) moves the RBS check logic into `new.rewrite` so the tests would again need a sigil (unless we wanted to drop the sigil check and just look for RBS?) I could probably make a test helper to insert sigils to contents but I wonder if that would be more confusing. What do you think?

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
paracycle: @dejmedus No problem. I am glad to hear it is helpful.

As for `initialize`, there is nothing magic about that method, and returning early doesn't change behaviour in any way different than in other methods. Even an empty `initialize` method still initializes the object, so you can add any logic in `initialize` that you want.

paracycle (approved):

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
dejmedus: > factored is 1.01x faster than flat

Interesting!

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -7,11 +7,42 @@ module Translate
...
+        ].freeze #: Array[String]
+        RBS_REWRITE_PATTERN = Regexp.union(["#:", "#|", *RBS_ANNOTATION_MARKERS]).freeze #: Regexp
+        private_constant :RBS_ANNOTATION_MARKERS, :RBS_REWRITE_PATTERN
+
         #: (String, file: String, ?max_line_length: Integer?) -> void
         def initialize(ruby_contents, file:, max_line_length: nil)
-          super(ruby_contents, file: file)
+          @ruby_contents = ruby_contents
+          if contains_rbs_syntax?
+            super(ruby_contents, file: file)
```
Morriar: So if `contains_rbs_syntax?` returns false we still initialize the object but do not call `super`. This means instance methods after that are responsible of knowing which state the instance is in. It's bad design.

Why do we even instantiate the `Translator` if we don't need to rewrite? This check should be made earlier.

Let's extract this to a `maybe_rewrite` singleton method.

Morriar (changes requested):
design is wrong, see comment

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
dejmedus: If I'm understanding correctly, I believe this is what we [originally had here](https://github.com/Shopify/spoom/pull/917/changes/b3862c1da8806e4a437921ce295e4c2cfa5da165#diff-3a035aa7b7631b6ac64e7316797385dd56355646b95ce2018576021bd72ff712R30) but it was [discussed](https://github.com/Shopify/spoom/pull/917#discussion_r3236978722) that we could prevent needing to pass around values [if the logic was moved inside](https://github.com/Shopify/spoom/pull/917#discussion_r3237452198)

Morriar: The conditional call to `super` is a code smell, the fact that we can't test translation without the sigils is another one.

The main problems:
- Fragile: forgetting which condition triggers super leads to half-initialized objects. Future maintainers have to reason about two different initialization paths.
- Violates Liskov: subclass instances behave differently depending on whether the parent was initialized, which breaks substitutability.
- Hard to test: you need to cover both branches, and bugs in the "no super" path tend to surface late as `nil` errors.

Instead of short-circuiting the `super` call we should just short-circuit the class instantiation altogether since we don't need it.

dejmedus: Ah, I understand, thank you. I dropped the commit and now are [exiting before initialization](https://github.com/Shopify/spoom/pull/917/changes#diff-3a035aa7b7631b6ac64e7316797385dd56355646b95ce2018576021bd72ff712R31) and [calling `new.rewrite`](https://github.com/Shopify/spoom/pull/917/changes#diff-f9a0f81ea7a119caba847b7c157e0e2eb0e7d9a0538d2a8a8dab65b1f00502b5R820) in tests

Morriar (approved):
Thanks! Sorry for the back and forth about the design.

amomchilov (approved):
shipit

PR #920, opened 2026-05-13
Add `rubocop-minitest`

amomchilov:
And turn on `Minitest/AssertPredicate`, matching [Ruby LSP](https://github.com/Shopify/ruby-lsp/blob/4f15ff61cd26ec8c7f47490cc86c09ee0792cd1f/.rubocop.yml#L34-L35)

### DIFF
--- a/test/spoom/context/git_test.rb
+++ b/test/spoom/context/git_test.rb
@@ -176,7 +176,7 @@ module Spoom
         assert(sha.size < 40)
 
         sha = T.must(context.git_last_commit(short_sha: false)).sha
-        assert(sha.size == 40)
+        assert_equal(sha.size, 40)
 
         last_commit = context.git_last_commit
         assert_equal(time.to_i, last_commit&.timestamp)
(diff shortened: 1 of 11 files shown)


test/spoom/context/git_test.rb
```diff
@@ -176,7 +176,7 @@ def test_context_git_last_commit
         assert(sha.size < 40)
 
         sha = T.must(context.git_last_commit(short_sha: false)).sha
-        assert(sha.size == 40)
+        assert_equal(sha.size, 40)
```
Morriar: The auto-correct put the arguments in the wrong order. Convention is `assert_equal(expected, actual)`, and the line right below follows it correctly.

```suggestion
        assert_equal(40, sha.size)
```

amomchilov: Ah [`Minitest/LiteralAsActualArgument`](https://docs.rubocop.org/rubocop-minitest/latest/cops_minitest.html#minitestliteralasactualargument) is not enabled by default yet (it's `pending`). I enabled it and auto-corrected this.

KaanOzkan (approved):

PR #921, opened 2026-05-13
Add `overloads_strategy` option to `RBSCommentsToSorbetSigs`

### LINKED ISSUE #913: `RBSCommentsToSorbetSigs` emits two consecutive `sig` blocks for overloaded `#:` annotations, which `sorbet-runtime` rejects at load time
## Summary

`Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs` translates each `#:` line preceding a method into a separate `sig do … end` block. For methods that use **overloaded** RBS annotations (multiple `#:` lines on a single method — a valid and common RBS pattern, e.g. for the with-block / without-block split on enumerable `each` methods), this produces two `sig` calls back-to-back. When the translated source is then loaded under `sorbet-runtime` — which is what tapioca's [`tapioca/rbs/rewriter.rb`](https://github.com/Shopify/tapioca/blob/v0.19.1/lib/tapioca/rbs/rewriter.rb) does via `RequireHooks.source_transform` during `bin/tapioca gem` — `T::Private::Methods._declare_sig_internal` raises:

```
RuntimeError: You called sig twice without declaring a method in between
```

…and gem load aborts. This makes `bin/tapioca gem <name>` fail for any gem that uses overloaded `#:` annotations. The one I hit it on is [`herb`](https://github.com/marcoroth/herb) (see [`Herb::DiffResult#each`, lines 17-18](https://github.com/marcoroth/herb/blob/v0.10.1/lib/herb/diff_result.rb#L17-L18)).

## Minimal reproducer

Standalone — only `spoom` and `sorbet-runtime`, no tapioca or herb required:

```ruby
require "spoom"
require "sorbet-runtime"

source = <<~RUBY
  # typed: true
  class Foo
    #: () { (Integer) -> void } -> void
    #: () -> Enumerator[Integer, void]
    def each(&block)
      return [1, 2, 3].each unless block
      [1, 2, 3].each(&block)
    end
  end
RUBY

translated = Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs(source, file: "(repro)")
puts translated

Module.include(T::Sig)
Object.class_eval(translated, "(repro-translated)")
```

### Spoom's output

```ruby
# typed: true
class Foo
  sig { params(block: ::T.proc.params(arg0: Integer).void).void }
  sig { returns(::T::Enumerator[Integer, void]) }
  def each(&block)
    return [1, 2, 3].each unless block
    [1, 2, 3].each(&block)
  end
end
```

### Runtime failure

```
RuntimeError: You called sig twice without declaring a method in between
  from sorbet-runtime/lib/types/private/methods/_methods.rb:55:in '_declare_sig_internal'
  from sorbet-runtime/lib/types/private/methods/_methods.rb:40:in 'declare_sig'
  from sorbet-runtime/lib/types/sig.rb:28:in 'T::Sig#sig'
  from (repro-translated):4
```

## Why this matters

Tapioca's `rbs/rewriter.rb` installs the `RequireHooks` transform globally for `**/*.rb` during gem RBI generation. Any gem in the bundle that uses overloaded `#:` annotations will abort tapioca's gem load. The blast radius scales with RBS adoption — and overloads are exactly the case where RBS is most expressive vs. Sorbet's single-sig syntax.

Concrete real-world example: I'm trying to migrate [packwerk](https://github.com/Shopify/packwerk) from `better_html` (deprecated) to `herb`, and `bin/tapioca gem herb` fails on load with the trace above. [The relevant herb source is one block-overloaded `each`](https://github.com/marcoroth/herb/blob/v0.10.1/lib/herb/diff_result.rb#L17-L18) — a textbook RBS overload pattern.

## Possible remediations

A few options for `RBSCommentsToSorbetSigs`, roughly in order of "amount of work":

1. **Emit only the last sig** when a method has multiple `#:` annotations. This matches Sorbet's existing "last sig wins" behavior for overloaded sigs (see [Shopify/tapioca#1789](https://github.com/Shopify/tapioca/issues/1789)). Loses overload fidelity in the rewritten source but unblocks gem load.
2. **Raise `Spoom::Sorbet::Translate::Error`** when overloads are detected. Tapioca's rewriter already rescues this and falls back to the original source — so the `#:` lines would stay as inert comments and the gem would load without runtime sigs. Tapioca's generator would produce an untyped (but otherwise valid) RBI.
3. **Collapse the overloads** into a single `sig` with appropriately-typed params/returns (e.g. accept `T.nilable(T.proc...)` for the optional block; return type becomes `T.any(T::Enumerator[...], void)`). Closest in spirit to the original RBS but a substantial implementation lift.

(1) seems like the smallest viable fix and would unblock everyone immediately; (2) is even smaller but produces less useful RBIs. Happy to send a PR for whichever route you'd prefer.

## Environment

- spoom 1.7.13
- sorbet-runtime 0.6.13184
- tapioca 0.19.1 (downstream)
- Ruby 3.4.9

Morriar:
## Summary

When a method has multiple `#:` annotations (RBS overloads), `RBSCommentsToSorbetSigs` emits multiple consecutive `sig` blocks. `sorbet-runtime` rejects this at load time with:

```
RuntimeError: You called sig twice without declaring a method in between
```

This breaks `tapioca gem` for any gem using overloaded RBS annotations (e.g. the common `each` with-block/without-block pattern).

This PR adds an `overloads_strategy` keyword parameter to `rbs_comments_to_sorbet_sigs` that accepts one of three symbols:

* `:translate_all` (default) -- current behavior, translates all overloads into separate `sig` blocks. No breaking change.
* `:translate_last` -- keeps only the last signature (closest to the method def, matching Sorbet's "last sig wins" semantics) and removes the others.
* `:raise` -- raises `Translate::Error`, allowing callers like tapioca to rescue and fall back to the original source.

Fixes #913

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
@@ -20,25 +20,33 @@ module Spoom
         RBS_REWRITE_PATTERN = Regexp.union(["#:", "#|", *RBS_ANNOTATION_MARKERS]).freeze #: Regexp
         private_constant :RBS_ANNOTATION_MARKERS, :RBS_REWRITE_PATTERN
 
+        ALLOWED_OVERLOAD_STRATEGIES = [:translate_all, :translate_last, :raise].freeze #: Array[Symbol]
+
         class << self
           #: (String source) -> bool
           def contains_rbs_syntax?(source)
             Sigils.contains_valid_sigil?(source) && source.match?(RBS_REWRITE_PATTERN)
           end
 
-          #: (String ruby_contents, file: String, ?max_line_length: Integer?) -> String
-          def rewrite_if_needed(ruby_contents, file:, max_line_length: nil)
+          #: (String ruby_contents, file: String, ?max_line_length: Integer?, ?overloads_strategy: Symbol) -> String
+          def rewrite_if_needed(ruby_contents, file:, max_line_length: nil, overloads_strategy: :translate_all)
             return ruby_contents unless contains_rbs_syntax?(ruby_contents)
 
-            new(ruby_contents, file:, max_line_length:).rewrite
+            new(ruby_contents, file:, max_line_length:, overloads_strategy:).rewrite
           end
         end
 
-        #: (String, file: String, ?max_line_length: Integer?) -> void
-        def initialize(ruby_contents, file:, max_line_length: nil)
+        #: (String, file: String, ?max_line_length: Integer?, ?overloads_strategy: Symbol) -> void
+        def initialize(ruby_contents, file:, max_line_length: nil, overloads_strategy: :translate_all)
           super(ruby_contents, file: file)
 
+          unless ALLOWED_OVERLOAD_STRATEGIES.include?(overloads_strategy)
+            raise ArgumentError, "Unknown overloads_strategy: #{overloads_strategy.inspect}. " \
+              "Must be one of: #{ALLOWED_OVERLOAD_STRATEGIES.map(&:inspect).join(", ")}"
+          end
+
           @max_line_length = max_line_length
+          @overloads_strategy = overloads_strategy
         end
 
         # @override
@@ -107,7 +115,13 @@ module Spoom
 
           return if comments.signatures.empty?
 
-          comments.signatures.each do |signature|
+          signatures = apply_overloads_strategy(
+            comments.signatures,
+            method_name: node.message.to_s,
+            location: "#{@file}:#{node.location.start_line}",
+          )
+
+          signatures.each do |signature|
             attr_type = ::RBS::Parser.parse_type(signature.string)
             sig = RBI::Sig.new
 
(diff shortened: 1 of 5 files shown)


lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -153,6 +165,26 @@ def rewrite_def(def_node, comments)
...
 
+        #: (Array[RBS::Signature]) -> Array[RBS::Signature]
+        def apply_overloads_strategy(signatures)
+          return signatures if signatures.size <= 1
+
+          case @overloads_strategy
+          when :translate_all
+            signatures
+          when :translate_last
+            kept = signatures.first #: as RBS::Signature
```
paracycle: It is strange that we keep `first` in a branch that says `..._last`. Do we get the signatures in reverse order? It might be good to add a note here to that effect.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -153,6 +165,26 @@ def rewrite_def(def_node, comments)
...
+            signatures.drop(1).each do |signature|
+              from = adjust_to_line_start(signature.location.start_offset)
+              to = adjust_to_line_end(signature.location.end_offset)
+              @rewriter << Source::Delete.new(from, to)
+            end
+            [kept]
+          else # :raise
+            raise Error, "Method has multiple overloaded signatures"
```
amomchilov: Can we give more context on which method, where? This could be hard to find.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
Morriar: I changed the helper to collect the signatures in the code order so we can use `last` and be less strange.

paracycle (approved):

PR #929, opened 2026-05-30
Fix dead code false positives for public_send, nested validates, and GraphQL macros

corsonknowles:
## Summary

Fixes three sources of dead code false positives:

### 1. `public_send` support in Ruby plugin

`send`, `__send__`, and `try` already create method references for their first symbol argument, but `public_send` was missing. This caused methods invoked via `obj.public_send(:method_name)` to be incorrectly flagged as dead.

### 2. Nested validation option hashes in ActiveModel plugin

`validates :attr, inclusion: { in: :method_name }` now correctly references `:method_name` as a method. Previously, only the top-level validation key was processed (creating a constant reference for the validator class), but symbol values inside nested option hashes were ignored.

### 3. GraphQL `argument`/`builds`/`prepare:` support

- `argument :name, prepare: :method_name` now references `:method_name`
- `builds :thing` now references `build_thing` (graphql-ruby convention)
- `argument :name, method: :method_name` and `field :name, method: :method_name` now reference the method (complementing the existing `resolver_method:` support)

## Context

These patterns were discovered while running `spoom deadcode` on a large Rails monolith (46K files). Of 77 candidates investigated across 10 packs, 39 were false positives — the majority from these three categories plus app-specific patterns (AASM, routes, ArDoc YAML) that are better suited for custom plugins.

## Test plan

Added minitest cases for each fix following existing plugin test patterns.

### DIFF
--- a/lib/spoom/deadcode/plugins/active_model.rb
+++ b/lib/spoom/deadcode/plugins/active_model.rb
@@ -30,6 +30,18 @@ module Spoom
                 @index.reference_method(value.slice.delete_prefix(":"), send.location) if value
               else
                 @index.reference_constant(camelize(key), send.location)
+
+                if value.is_a?(Prism::HashNode)
+                  value.elements.each do |assoc|
+                    next unless assoc.is_a?(Prism::AssocNode)
+
+                    assoc_key = assoc.key.slice.delete_suffix(":")
+                    next unless assoc_key == "in" || assoc_key == "with"
+                    next unless assoc.value.is_a?(Prism::SymbolNode)
+
+                    @index.reference_method(assoc.value.unescaped, send.location)
+                  end
+                end
               end
             end
           when "validates_with"
--- a/lib/spoom/deadcode/plugins/active_model.rb
+++ b/lib/spoom/deadcode/plugins/active_model.rb
@@ -37,9 +37,10 @@ module Spoom
 
                     assoc_key = assoc.key.slice.delete_suffix(":")
                     next unless assoc_key == "in" || assoc_key == "with"
-                    next unless assoc.value.is_a?(Prism::SymbolNode)
+                    value = assoc.value
+                    next unless value.is_a?(Prism::SymbolNode)
 
-                    @index.reference_method(assoc.value.unescaped, send.location)
+                    @index.reference_method(value.unescaped, send.location)
                   end
                 end
               end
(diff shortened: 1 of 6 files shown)


test/spoom/deadcode/plugins/active_model_test.rb
```diff
@@ -83,6 +83,27 @@ class MyValidator7; end
...
+              def max_value; end
+              def pattern_method; end
+              def dead; end
+            end
+          RB
+
+          index = index_with_plugins
+          assert_alive(index, "allowed_values")
+          assert_alive(index, "pattern_method")
+          assert_dead(index, "max_value")
```
amomchilov, line 103: Why is `max_value` dead?

amomchilov:
Hi Dave, thanks for contributing these. Could you please split them into 3 separate PRs?

I approved the CI run in the meantime, so you can iterate on any errors that might come up

test/spoom/deadcode/plugins/active_model_test.rb
corsonknowles: It's not dead code, the fix is just too limited. I'll pull the list and apply all the Rails methods that invoke symbolic procs in this context.

lib/spoom/deadcode/plugins/active_model.rb
```diff
@@ -30,6 +30,18 @@ def on_send(send)
                 @index.reference_method(value.slice.delete_prefix(":"), send.location) if value
               else
                 @index.reference_constant(camelize(key), send.location)
+
+                if value.is_a?(Prism::HashNode)
+                  value.elements.each do |assoc|
```
amomchilov, line 35: Can you use `each_arg_assoc` here?

lib/spoom/deadcode/plugins/active_model.rb
```diff
@@ -30,6 +30,18 @@ def on_send(send)
...
+                    next unless assoc.is_a?(Prism::AssocNode)
+
+                    assoc_key = assoc.key.slice.delete_suffix(":")
+                    next unless assoc_key == "in" || assoc_key == "with"
+                    next unless assoc.value.is_a?(Prism::SymbolNode)
+
+                    @index.reference_method(assoc.value.unescaped, send.location)
```
amomchilov: `assoc.value` is a method call, with no (statically proveable) guarantee that it returns the same thing from one call to the next.

To solve the type-checking issue, call it only once:


```suggestion
                    value = assoc.value
                    next unless value.is_a?(Prism::SymbolNode)

                    @index.reference_method(value.unescaped, send.location)
```

corsonknowles:
Thanks!

Closed in favor of:
* https://github.com/Shopify/spoom/pull/930

* https://github.com/Shopify/spoom/pull/931 

* https://github.com/Shopify/spoom/pull/932

PR #931, opened 2026-05-30
Reference symbol method names in nested validates option hashes

corsonknowles:
## Problem

`validates :attr, inclusion: { in: :allowed_values }` incorrectly flags `allowed_values` as dead code. The ActiveModel plugin processes top-level keyword args but does not descend into nested option hashes where Rails validators accept symbol method references.

## Fix

When processing `validates`/`validates!` keyword args, if a value is a HashNode, the plugin now iterates its assocs and references SymbolNode values as methods for known method-reference keys:

- **inclusion/exclusion:** `in:`
- **format:** `with:`
- **numericality:** `less_than:`, `greater_than:`, `less_than_or_equal_to:`, `greater_than_or_equal_to:`, `equal_to:`, `other_than:`

All of these Rails validation options accept a symbol that is called as a method on the model instance at validation time.

## Test plan

Added minitest cases covering all nested option keys with symbol values.

### DIFF
--- a/lib/spoom/deadcode/plugins/active_model.rb
+++ b/lib/spoom/deadcode/plugins/active_model.rb
@@ -69,9 +69,11 @@ module Spoom
 
             nested_key = assoc.key.slice.delete_suffix(":")
             next unless NESTED_METHOD_REFERENCE_KEYS.include?(nested_key)
-            next unless assoc.value.is_a?(Prism::SymbolNode)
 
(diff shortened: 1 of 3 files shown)


amomchilov (approved):

corsonknowles:
Thanks for all the help @amomchilov ! I think this and the other one just need CI to be allowed to run

PR #932, opened 2026-05-30
Add GraphQL `argument`, `builds`, and `prepare` support to dead code plugin

corsonknowles:
## Problem

The GraphQL plugin only handles `field :name` (and `resolver_method:`). Methods referenced through other graphql-ruby DSL patterns are incorrectly flagged as dead:

- `argument :name, prepare: :method` — the prepare method is called by graphql-ruby when processing the input
- `builds :thing` — graphql-ruby convention that calls `build_thing` on the mutation
- `argument :name, method: :method` — custom method for argument resolution

## Fix

Extended the GraphQL plugin's `on_send` to handle `argument` (with `prepare:` and `method:` keywords) and `builds` (with convention-based `build_#{name}` method reference).

## Test plan

Added minitest cases for each new pattern.

### DIFF
--- a/lib/spoom/deadcode/plugins/graphql.rb
+++ b/lib/spoom/deadcode/plugins/graphql.rb
@@ -22,22 +22,59 @@ module Spoom
           "unsubscribed",
         )
 
+        FIELD_SYMBOL_OPTION_KEYS = ["resolver_method", "method"].freeze #: Array[String]
+        ARGUMENT_SYMBOL_OPTION_KEYS = ["prepare", "method"].freeze #: Array[String]
+
         # @override
         #: (Send send) -> void
         def on_send(send)
-          return unless send.recv.nil? && send.name == "field"
+          return unless send.recv.nil?
 
+          case send.name
+          when "field"
+            on_field(send)
+          when "argument"
+            on_argument(send)
+          when "builds"
+            on_builds(send)
+          end
+        end
+
+        private
+
+        #: (Send send) -> void
+        def on_field(send)
           arg = send.args.first
           return unless arg.is_a?(Prism::SymbolNode)
 
           @index.reference_method(arg.unescaped, send.location)
 
           send.each_arg_assoc do |key, value|
-            key = key.slice.delete_suffix(":")
-            next unless key == "resolver_method"
+            key_name = key.slice.delete_suffix(":")
+            next unless FIELD_SYMBOL_OPTION_KEYS.include?(key_name)
             next unless value
 
             @index.reference_method(value.slice.delete_prefix(":"), send.location)
+          end
+        end
+
+        #: (Send send) -> void
+        def on_argument(send)
+          send.each_arg_assoc do |key, value|
+            key_name = key.slice.delete_suffix(":")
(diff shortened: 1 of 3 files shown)


lib/spoom/deadcode/plugins/graphql.rb
```diff
@@ -40,6 +54,26 @@ def on_send(send)
             @index.reference_method(value.slice.delete_prefix(":"), send.location)
           end
         end
+
+        #: (Send send) -> void
+        def on_argument(send)
+          send.each_arg_assoc do |key, value|
+            key = key.slice.delete_suffix(":")
```
amomchilov: I know you just matched this to the code above, but this has a latent bug: it would be wrong for e.g. `:"hello world"` or `%s[foo]`.

Should have been `key = key.value`. Leave it as-is in this PR, and I'll try to fix both next week.

Or, if you're feeling industrious, you're you're welcome to contribute another PR for a fix if you'd like :)

lib/spoom/deadcode/plugins/graphql.rb
```diff
@@ -40,6 +54,26 @@ def on_send(send)
...
+            @index.reference_method(value.slice.delete_prefix(":"), send.location)
+          end
+        end
+
+        #: (Send send) -> void
+        def on_builds(send)
+          send.args.each do |arg|
+            next unless arg.is_a?(Prism::SymbolNode)
+
+            @index.reference_method("build_#{arg.unescaped}", send.location)
```
amomchilov, line 77: TODO for me to look up with fresh eyes: What's the difference between `Prism::SymbolNode#unescaped` and `#value`

lib/spoom/deadcode/plugins/ruby.rb
```diff
@@ -24,7 +24,7 @@ def on_send(send)
           case send.name
           when "const_defined?", "const_get", "const_source_location"
             reference_symbol_as_constant(send, T.must(send.args.first))
-          when "send", "__send__", "try"
+          when "send", "__send__", "public_send", "try"
```
amomchilov: wrong PR

lib/spoom/deadcode/plugins/graphql.rb
amomchilov, line 1: uh oh, lost the test changes in this PR

lib/spoom/deadcode/plugins/graphql.rb
amomchilov: Opened an issue to track this: https://github.com/Shopify/spoom/issues/933

corsonknowles: Great, if I understood that correctly, this should help:

* https://github.com/Shopify/spoom/pull/934

amomchilov (approved):

amomchilov:
Thanks!

lib/spoom/deadcode/plugins/graphql.rb
amomchilov: Documented the difference in https://github.com/Shopify/spoom/issues/933

PR #934, opened 2026-06-01
GraphQL plugin: parse argument/field Symbols via Prism `#value`

### LINKED ISSUE #933: Incorrect Symbol parsing via `.slice.delete_prefix(":")`
## Problem

This snippet only works for regular `:foo` symbols, but not the other variations:

```ruby
require "prism"
def parse(s) = Prism.parse(s).value.statements.body

puts(parse(<<~RUBY).map { |sym_node| sym_node.slice.delete_prefix(":") })
  :works
  %s[weird delimiters don't get cut out]
  :"interpolated #{123} symbols stayed quoted"
RUBY
```

Prints:

```
works
%s[weird delimiters don't get cut out]
"interpolated 123 symbols stayed quoted"
```

`delete_suffix(":")` for kwarg symbol names has the same problem.

## Usages

https://github.com/search?q=repo%3AShopify%2Fspoom+%2Fdelete_%28prefix%7Csuffix%29%5C%28%22%3A%22%5C%29%2F&type=code

## Solution

Use `#unescaped` (not `#value`):

```ruby
irb(main):001> node = Prism.parse(':"foo\n"').value.statements.body.first
irb(main):002> node.value
=> "foo\\n" # (verbatim from source, with literal backslash and n)
irb(main):003> node.unescaped # => "foo\n"
=> "foo\n"  # (real new line)
```

> For a plain `:foo` they're identical strings.
> Rule of thumb: use `#unescaped` when you want the Symbol the program will actually create (`node.unescaped.to_sym`). Use `#value` when you want to know what the user typed.

So `unescaped` is what we want to use for mapping to runtime values like method names.

corsonknowles:
## Problem
Fixes #933 for the GraphQL plugin. The plugin parsed kwarg option keys and symbol values with `slice.delete_suffix(":")` / `slice.delete_prefix(":")`. As @amomchilov noted on #932, that's wrong for quoted/escaped symbols — e.g. `:"weird name"` has `slice == ":\"weird name\""`, so `delete_prefix(":")` leaves `"\"weird name\""` (quotes intact) and the referenced method name is mangled, causing the method to be reported dead.

## Fix
Use `Prism::SymbolNode#value` for both the option key and the symbol value, and guard on `Prism::SymbolNode` so non-symbol keys/values (including interpolated symbols, which are `InterpolatedSymbolNode`) are skipped:

```ruby
send.each_arg_assoc do |key, value|
  next unless key.is_a?(Prism::SymbolNode)
  next unless ARGUMENT_SYMBOL_OPTION_KEYS.include?(key.value)
  next unless value.is_a?(Prism::SymbolNode)

  @index.reference_method(value.value, send.location)
end
```

Verified with Prism that `#value` returns `"weird name"` where `slice.delete_prefix(":")` returns `"\"weird name\""`.

## Scope
Limited to `lib/spoom/deadcode/plugins/graphql.rb` (the `argument`/`field` handling added in #932). Other `delete_(prefix|suffix)(":")` usages tracked by #933 are intentionally left for a separate change.

## Stacked on #932
This is **stacked on #932** (`graphql-argument-builds-prepare`). Until #932 merges, this PR's diff also shows #932's commits; please merge **after** #932 (or rebase onto `main` once #932 lands).

## Test plan
Added `test_alive_argument_prepare_with_quoted_symbol` and `test_alive_field_method_with_quoted_symbol` to `graphql_test.rb`, asserting a quoted-symbol `prepare:`/`method:` value keeps the real method alive.

Made with Cursor


amomchilov:
Solved more generally in #938

Thanks!

PR #935, opened 2026-06-01
Fix invalid sig generated for anonymous block param (&)

### LINKED ISSUE #928: RBS Rewriting produces invalid Ruby for anonymous block parameters
## Description

`Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs` generates invalid Ruby when translating an RBS block type annotation on a method that uses Ruby 3.1+ anonymous block forwarding (`&`). The generated `sig` contains `&block:` inside `params()`, which is a syntax error. Named block parameters (`&block`) are translated correctly.

## Reproduction

This code:

```ruby
require "spoom"

source = <<~RUBY
  # typed: true
  class Foo
    #: (String) ?{ (String) -> void } -> String
    def bar(request, &); end
  end
RUBY

rewritten = Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs(source, file: "test.rb")
puts rewritten
RubyVM::InstructionSequence.compile(rewritten)
```

### Produces

```
# typed: true
class Foo
  sig { params(request: String, &block: ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
  def bar(request, &); end
end
(shopify):36:in 'RubyVM::InstructionSequence.compile': <compiled>:3: syntax errors found (SyntaxError)
  1 | # typed: true
  2 | class Foo
> 3 | ... : ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
    |     ^ unexpected ':', ignoring it
    |     ^ unexpected ':', expecting end-of-input
    |     ^ unexpected ':'; expected a `)` to close the arguments
    |                                                      ^ unexpected ')', ignoring it
    |                                                      ^ unexpected ')', expecting end-of-input
    |                                                       ^ unexpected '.', ignoring it
  4 |   def bar(request, &); end
  5 | end
```

### Expected

```
# typed: true
class Foo
  sig { params(request: String, block: ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
  def bar(request, &); end
end
=> <RubyVM::InstructionSequence:<compiled>@<compiled>:1>
```

## Environment

- Spoom 1.7.15
- Ruby 3.4.9
- Tapioca 0.19.1

prattik-wav:
Fixes #928

## Problem

`rbs_comments_to_sorbet_sigs` generates invalid Ruby when a method uses Ruby 3.1+ anonymous block forwarding (`def foo(&); end`). The generated sig contains `&block:` inside `params()`, which is a syntax error:

```ruby
# Input
#: (String) ?{ (String) -> void } -> String
def bar(request, &); end

# Generated (broken)
sig { params(request: String, &block: ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
def bar(request, &); end
```

Named block params (`&block`) were already translated correctly.

## Root Cause

`Prism::BlockParameterNode#name` is `nil` for anonymous `&`. The `RBI::RBS::MethodTypeTranslator` doesn't distinguish this case and falls back to the default name `&block`, which renders as `&block:` inside `params()` — a syntax error because `&` is not valid in a keyword argument position.

## Fix

After `translator.result` produces the sig, check whether the original `def_node` has an anonymous block parameter (a `BlockParameterNode` with `nil` name). If so, strip the leading `&` from any block param name in the generated sig, producing the valid `block:` form.

```
# Generated (fixed)
sig { params(request: String, block: ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
def bar(request, &); end
```

## Testing

- Added `test_translate_to_rbi_anonymous_block_param` — covers the broken case from the issue
- Added `test_translate_to_rbi_named_block_param_unchanged` — regression guard confirming named `&block` still works
- Verified the fixed output compiles cleanly with `RubyVM::InstructionSequence.compile`

Fixes #928


test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -890,6 +890,45 @@ def rbs_comments_to_sorbet_sigs(ruby_contents, max_line_length: nil, overloads_s
...
+              #: (String) ?{ (String) -> void } -> String
+              def bar(request, &); end
+            end
+          RUBY
+
+          assert_equal(<<~RUBY, res)
+            # typed: true
+            class Foo
+              sig { params(request: String, block: ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
+              def bar(request, &); end
```
paracycle, line 907: I don't think this generated code is correct either: https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20%7B%20params%28request%3A%20String%2C%20block%3A%20%3A%3AT.nilable%28%3A%3AT.proc.params%28arg0%3A%20String%29.void%29%29.void%20%7D%0A%20%20def%20bar%28request%2C%20%26%29%3B%20nil%3B%20end%0Aend

I think it should be:

```suggestion
              sig { params(request: String, "&": ::T.nilable(::T.proc.params(arg0: String).void)).returns(String) }
              def bar(request, &); end
```
as you can see working here: https://sorbet.run/#%23%20typed%3A%20true%0Aclass%20Foo%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20do%20%0A%20%20%20%20type_parameters%28%3AR%29%0A%20%20%20%20%20%20.params%28%22%26%22%3A%20%3A%3AT.proc.returns%28T.type_parameter%28%3AR%29%29%29%0A%20%20%20%20%20%20.returns%28T.type_parameter%28%3AR%29%29%0A%20%20end%0A%20%20def%20bar%28%26%29%3B%20T.unsafe%28nil%29%20end%0Aend%0A%0Afoo%20%3D%20Foo.new%0Aresult%20%3D%20foo.bar%20do%20%7Carg0%7C%0A%20%20%3Asymbol%0Aend%0A%0AT.reveal_type%28result%29

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -184,6 +184,16 @@ def rewrite_def(def_node, comments)
...
 
+            # When the method uses Ruby 3.1+ anonymous block forwarding (`&` with no name),
+            # the RBI translator names the block param `&block`, producing `&block:` inside
+            # `params()` which is a syntax error. Detect that case and rename it to `block`
+            # (without the `&`) so the generated sig is valid Ruby.
+            if anonymous_block_param?(def_node)
+              sig.params.each do |param|
+                param.name = param.name.delete_prefix("&") if param.name.start_with?("&")
+              end
+            end
```
paracycle, line 195: I am not sure if this is the correct place to fix this problem. We need to solve it at the source.

paracycle (changes requested):

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -890,6 +890,45 @@ def rbs_comments_to_sorbet_sigs(ruby_contents, max_line_length: nil, overloads_s
...
+            end
+          RUBY
+
+          # Must also be valid Ruby
+          assert RubyVM::InstructionSequence.compile(res)
```
amomchilov, line 912: This is overkill, we don't need to waste time compiling it. Perhaps `assert Prism.parse_success?(res)` at most, but we match the exact expected source we want, so there's no need to confirm this here


```suggestion
```

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
amomchilov: Yeah the correct place to fix this is in the RBI gem that Spoom uses for this translation:

https://github.com/Shopify/rbi/blob/4a4f30b9ca743f7723ce451dc06a9b7cf62a42b0/lib/rbi/rbs/method_type_translator.rb#L46

PR #939, opened 2026-06-05
Extract `#assert_rewrites_rbs` helper

amomchilov:
## Background

As part of #787, I'm building a more comprehensive test suite that will ensure that the RBS-to-Sorbet translator preserves line numbering.

Achieving correct line numbers while still keeping the same meaning will require the translator to produce some [pretty funky code](https://github.com/Shopify/spoom/issues/787#issuecomment-4595627848), which is great for machines, but sucks for humans. So we'll want to support both:
1. a more human-readable format that limits line lengths, has reasonable new lines for clarity, etc.
    * this is what we already have
2. a line-matching format that "works" for machines, and preserves the correctness of line numbers in back traces, `__LINE__`, etc.
    * this is what I'm working on building

## This PR

Refactors our existing tests into a new `#assert_rewrites_rbs` helper. Currently it just has one input and one expected output.

In a future PR, I'll be adding a second expected output for each test case. WIP, but you can get an idea here: https://github.com/Shopify/spoom/commit/658304dc509bde0ab7adab14965689dda56223df

Very easy to review [if you hide whitespace changes](https://github.com/Shopify/spoom/pull/939/changes?w=1).

## Alternatives considered

Make a dedicated test class for the line-matching rewriter. That way this test class' test cases have one input and one (human-readable) expected output, and the other class' test cases have one input and one (machine-readable) expected output.

I [tried that out](https://github.com/Shopify/spoom/blob/cc5fe85d8857357eb6a897a47275c4e1a32ca926/test/spoom/sorbet/translate/rbs_comments_to_line_matched_sorbet_sigs_test.rb), but it became obvious that the two suites would be hard to keep in sync, and we would surely have new additions in the future that are only tested in one mode or the other, rather than both.

### DIFF
--- a/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
+++ b/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
@@ -9,7 +9,10 @@ module Spoom
       class RBSCommentsToSorbetSigsTest < Minitest::Test
         def test_translate_to_rbi_empty
           contents = ""
-          assert_equal(contents, rbs_comments_to_sorbet_sigs(contents))
+          assert_rewrites_rbs(
+            from: contents,
+            to_pretty_format_for_humans: contents,
+          )
         end
 
         def test_translate_to_rbi_no_sigs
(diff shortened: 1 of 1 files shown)


test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -890,6 +970,32 @@ def rbs_comments_to_sorbet_sigs(ruby_contents, max_line_length: nil, overloads_s
...
           ).rewrite
         end
+
+        #: (
+        #|   from: String,
+        #|   to_pretty_format_for_humans: String,
+        #|   ?max_line_length: Integer?,
+        #|   ?overloads_strategy: Symbol
+        #|  ) -> String
+        def assert_rewrites_rbs(
```
KaanOzkan, line 980: Looks like there are 7 occurrences of tests where the `from` and `to_pretty_format_for_humans` is equivalent. Wdyt about adding a default value so that if they're equal the callsites won't have to specify the kwargs:

```rb
def assert_rewrites_rbs(
  from = nil,
  to: from,
  max_line_length: nil,
  overloads_strategy: :translate_all
)
```

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -890,6 +970,32 @@ def rbs_comments_to_sorbet_sigs(ruby_contents, max_line_length: nil, overloads_s
             overloads_strategy: overloads_strategy,
           ).rewrite
         end
+
+        #: (
+        #|   from: String,
+        #|   to_pretty_format_for_humans: String,
+        #|   ?max_line_length: Integer?,
+        #|   ?overloads_strategy: Symbol
+        #|  ) -> String
```
KaanOzkan:
```suggestion
        #|  ) -> void
```
Seems more idiomatic

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -890,6 +970,32 @@ def rbs_comments_to_sorbet_sigs(ruby_contents, max_line_length: nil, overloads_s
...
+        def assert_rewrites_rbs(
+          from:,
+          to_pretty_format_for_humans:,
+          max_line_length: nil,
+          overloads_strategy: :translate_all
+        )
+          source_with_rbs = from
+          expected_pretty_format = to_pretty_format_for_humans
+
+          begin # Validate the human-readable rewrite
```
KaanOzkan, line 989: Why is this wrapped in a `begin`?

amomchilov: Just for visual grouping

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
amomchilov: Spoke offline. Will following up with something similar, but as a separate method like `assert_rbs_rewrite_is_no_op`.

That way you don't have a call with a `from:` but not `to:`

amomchilov: Oh also, some things that are no-op today, might not be no-op in the new rewriter mode. I'll explain it in the PR when I open it.

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -6,10 +6,13 @@
 module Spoom
   module Sorbet
     module Translate
-      class RBSCommentsToSorbetSigsTest < Minitest::Test
+      class RBSCommentsToSigsTest < Minitest::Test
```
alexcrocha: Should we also rename `RBSCommentsToSorbetSigs` to `RBSCommentsToSigs` to match?

alexcrocha (approved):

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
alexcrocha: Felt weird at first sight, but it quickly grew up on me 👍

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
amomchilov: Unintentional rename from when I was merging two files. Undone!

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
amomchilov: For context, up the stack there will be two of these, running different sets of assertions/validations against the two different outputs: https://github.com/Shopify/spoom/blob/658304dc509bde0ab7adab14965689dda56223df/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb#L1035-L1066

PR #940, opened 2026-06-05
Line-matching rewriter

amomchilov:
This PR adds a line preserving rewriter from RBS to Sorbet sigs, which is helpful when validating the runtime so that backtraces don't get shifted by the translation.

The idea is to introduce a new format, which changes the way we translate to not always simply delete the RBS, but instead keep artifacts that will maintain the original line numbers.

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
@@ -100,6 +100,8 @@ module Spoom
               location: "#{@file}:#{node.location.start_line}",
             )
 
+            known_annotations = nil #: Array[Spoom::RBS::Annotation]?
+
             signatures.each do |signature|
               attr_type = ::RBS::Parser.parse_type(signature.string)
               sig = RBI::Sig.new
@@ -118,7 +120,7 @@ module Spoom
 
               sig.return_type = @type_translator.translate(attr_type)
 
-              apply_member_annotations(comments.method_annotations, sig)
+              known_annotations = apply_member_annotations(comments.method_annotations, sig)
 
               @rewriter << Source::Replace.new(
                 signature.location.start_offset,
@@ -128,6 +130,10 @@ module Spoom
             rescue ::RBS::ParsingError, ::RBI::Error
               # Ignore signatures with errors
               next
+            end
+
+            if known_annotations
+              rewrite_member_annotations(comments.method_annotations, known: known_annotations)
             end
           end
 
@@ -146,6 +152,8 @@ module Spoom
             builder.visit(def_node)
             rbi_node = builder.tree.nodes.first #: as RBI::Method
 
+            known_annotations = nil #: Array[Spoom::RBS::Annotation]?
+
             signatures.each do |signature|
               begin
                 method_type = ::RBS::Parser.parse_method_type(signature.string)
@@ -163,7 +171,7 @@ module Spoom
 
               sig = translator.result
 
-              apply_member_annotations(comments.method_annotations, sig)
+              known_annotations = apply_member_annotations(comments.method_annotations, sig)
 
               # Sorbet runtime doesn't support `sig` on `method_added` or
               # `singleton_method_added`, so we always use `without_runtime` for them.
@@ -176,6 +184,10 @@ module Spoom
                 signature.location.end_offset,
                 pad_out_line_count(of: sig.string(max_line_length: @max_line_length), to_height_of: signature),
               )
+            end
+
+            if known_annotations
+              rewrite_member_annotations(comments.method_annotations, known: known_annotations)
             end
           end
 
@@ -315,8 +327,10 @@ module Spoom
           #: (String type_member, parent_node: PrismTypes::anyScopeNode, insert_pos: Integer) -> void
           def insert_type_member(type_member, parent_node:, insert_pos:) = raise
 
-          #: (Array[Spoom::RBS::Annotation], RBI::Sig) -> void
+          #: (Array[Spoom::RBS::Annotation], RBI::Sig) -> Array[Spoom::RBS::Annotation]
           def apply_member_annotations(annotations, sig)
+            known = [] #: Array[Spoom::RBS::Annotation]
+
             annotations.each do |annotation|
               case annotation.string
               when "@abstract"
@@ -336,11 +350,23 @@ module Spoom
               when "@without_runtime"
                 sig.without_runtime = true
               else
-                rewrite_annotation(annotation, is_known: false)
                 next
               end
 
-              rewrite_annotation(annotation, is_known: true)
+              known << annotation
+            end
+
+            known
+          end
+
+          # Rewrites the member annotation comments in the source. Called once per method,
+          # regardless of how many overloaded signatures share the annotations, to avoid
+          # emitting duplicate markers.
+          #
+          #: (Array[Spoom::RBS::Annotation], known: Array[Spoom::RBS::Annotation]) -> void
+          def rewrite_member_annotations(annotations, known:)
+            annotations.each do |annotation|
+              rewrite_annotation(annotation, is_known: known.include?(annotation))
             end
           end
 
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
@@ -123,7 +123,7 @@ module Spoom
               @rewriter << Source::Replace.new(
                 signature.location.start_offset,
                 signature.location.end_offset,
-                sig.string(max_line_length: @max_line_length),
+                pad_out_line_count(of: sig.string(max_line_length: @max_line_length), to_height_of: signature),
               )
             rescue ::RBS::ParsingError, ::RBI::Error
               # Ignore signatures with errors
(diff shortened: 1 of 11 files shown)


amomchilov:
* **#940** <a href="https://app.graphite.com/github/pr/Shopify/spoom/940?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/spoom/940?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#954** <a href="https://app.graphite.com/github/pr/Shopify/spoom/954?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#955](https://github.com/Shopify/spoom/pull/955) <a href="https://app.graphite.com/github/pr/Shopify/spoom/955?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -329,11 +341,26 @@ def apply_member_annotations(annotations, sig)
                 sig.is_overridable = true
               when "@without_runtime"
                 sig.without_runtime = true
+              else
+                rewrite_annotation(annotation, is_known: false)
+                next
               end
+
+              rewrite_annotation(annotation, is_known: true)
```
KaanOzkan: LLM review pointed out that this is getting called per signature, so we could have a case of duplicate `RBS_REWRITTEN_ANNOTATION ` being inserted

```rb
# @override
#: (Integer) -> void
#: (String) -> void
def foo(x); end
```

would be translated into

```rb
# RBS_REWRITTEN_ANNOTATION: RBS_REWRITTEN_ANNOTATION: @override
```

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/line_matching_translator.rb
```diff
@@ -0,0 +1,109 @@
...
+    module Translate
+      module RBSCommentsToSorbetSigs
+        class LineMatchingTranslator < BaseTranslator
+          private
+
+          # Comments out the discarded overload
+          # @override
+          #: (RBS::Signature) -> void
+          def rewrite_discarded_overload(signature)
+            @rewriter << Source::Insert.new(signature.location.start_offset + 1, " RBS_DISCARDED_OVERLOAD")
```
KaanOzkan, line 18: Another LLM edge case find: This isn't considering multi line signatures and the RBS syntax sticks around.

```rb
#: (Integer,
#|  String) -> void
```

gets translated into

```rb
# RBS_DISCARDED_OVERLOAD: (Integer,
#|  String) -> void
```

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/options.rb
```diff
@@ -31,6 +31,14 @@ class << self
           end
         end
 
+        class LineMatchedRBIFormat < BaseRBIFormat # TODO: move to RBI gem
+          @default = new #: LineMatchedRBIFormat
+          class << self
+            #: LineMatchedRBIFormat
+            attr_reader :default
+          end
+        end
```
KaanOzkan, line 39: Will this be used similar to `HumanReadableRBIFormat` in lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb? Curious why it's not being triggered in this PR.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -309,6 +300,27 @@ def apply_class_annotations(node)
...
+          # @param is_known: true if this is an RBS annotation that we recognize
+          #                  false for some other `@`-prefixed thing, like a documentation `@param` tag.
+          # @abstract
+          #: (
+          #|   Spoom::RBS::Annotation,
+          #|   parent_node: Prism::AnyScopeNode,
+          #|   insert_pos: Integer,
+          #|   sorbet_replacement: String?
+          #|  ) -> void
+          def apply_class_annotation(annotation, parent_node:, insert_pos:, sorbet_replacement:) = raise
```
KaanOzkan, line 318: These are making this class harder to read but I'm not sure what we can do. One alternative I kinda like it to move these abstract methods to `HumanReadableTranslator` instead. Call sites in this class would then delegate to the appropriate "mode" that's being set, `@translation_mode.apply_class_anotation`. It wouldn't get rid of the abstract but make the intent more clear. Wdyt? Maybe as a future improvement to not hold up the PR.

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -1021,6 +1205,46 @@ def assert_rewrites_rbs(
...
+                The rewritten code does not match the expected line-matched format.
+
+                Validation errors:
+                #{validation_result.errors.map { |e| "- #{e}" }.join("\n")}
+
+                Expected line-matched format:
+                #{expected_line_matched_format}
+
+                Actual rewritten output:
+                #{rewritten_output}
```
KaanOzkan: The message talks about line-matched format but `rewritten_output` is human readable output at this stage.

amomchilov: Good catch.

I want the test expectation to be validated _before_ proceeding to the rewrite, so I just fixed this by deleting "actual rewritten output" from this error message.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/options.rb
amomchilov: Yes. It's not used yet because I didn't finish any of the nicer line-matching formatting in the parts the RBI gem is responsible for printing.

Instead, I just pad with as many blank lines as we need.

In a future iteration, this will be used to signal to the RBI gem that e.g. it should break a multi-line RBS sig into a multiline Sorbet sig.

vinistock: When we want to expose this from the CLI, yes we need to do that work. For now, we're only using it via the API, so it's enough for this PR.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
vinistock: @amomchilov I'll leave this one for your assessment. I agree that something along the lines of what Kaan suggested would make the code a bit easier to follow.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -315,8 +327,10 @@ def rewrite_type_params_signature(signature, type_params:) = raise
           #: (String type_member, parent_node: PrismTypes::anyScopeNode, insert_pos: Integer) -> void
           def insert_type_member(type_member, parent_node:, insert_pos:) = raise
 
-          #: (Array[Spoom::RBS::Annotation], RBI::Sig) -> void
+          #: (Array[Spoom::RBS::Annotation], RBI::Sig) -> Array[Spoom::RBS::Annotation]
           def apply_member_annotations(annotations, sig)
+            known = [] #: Array[Spoom::RBS::Annotation]
+
             annotations.each do |annotation|
```
amomchilov, line 334: I'll follow-up to tidy this with `Array#map`

amomchilov (commented):
Nice fix!

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
amomchilov: Same. Will follow-up.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/line_matching_translator.rb
KaanOzkan: @vinistock I think this is still failing? Although it's an edge case

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/line_matching_translator.rb
```diff
@@ -0,0 +1,102 @@
...
+            if sorbet_replacement
+              @rewriter << Source::Insert.new(insert_pos, "; #{sorbet_replacement}")
+            end
+          end
+
+          # @override
+          #: (Spoom::RBS::Signature, type_params: Array[::RBS::AST::TypeParam]) -> void
+          def rewrite_type_params_signature(signature, type_params:)
+            # Rewrite `#: [A, B]` into `# RBS_WRITTEN_ANNOTATION: [A, B]`
+            @rewriter << Source::Replace.new(
```
KaanOzkan, line 53: Similar to previous comment, multiline + annotation edge case is failing:

```rb
#: [A,
#| B]
```
is translated into

```rb
# RBS_WRITTEN_ANNOTATION: [A,
#| B]
```

KaanOzkan (approved):
There are still 2 edge cases (one is hidden in a [resolved comment](https://github.com/Shopify/spoom/pull/940#discussion_r3444496674)) but I think the implementation is good. If it's not too hard it'd be good to try to fix these too.

alexcrocha (approved):
Approving to unblock. We can fix forward any edge cases that come up

KaanOzkan (approved):
Thanks!

PR #942, opened 2026-06-07
Remove sigs attached to methods defined with modifier calls

dduugg:
### Motivation

When the deadcode remover removes a method defined as the argument of a modifier
call — e.g. `private def foo; end`, `private_class_method def self.foo; end`, or
`abstract def foo; end` — it removes the method but leaves the method's `sig`
(and any attached comments) behind:

```ruby
class Foo
  sig { params(value: Integer).void }
  private_class_method def self.bar(value)
    something
  end

  def baz; end
end
```

Removing `bar` currently produces:

```ruby
class Foo
  sig { params(value: Integer).void }

  def baz; end
end
```

The orphaned `sig` then attaches to the *next* method definition (`baz`). This
goes wrong in a few ways:

- It's a Sorbet error when the leftover sig doesn't match the following method —
  here `baz` has no `value` parameter.
- When the following method has its *own* sig, removal leaves two `sig`s with no
  method definition between them, which is also a Sorbet error.
- When the leftover sig happens to be compatible with the following method, it's
  arguably worse: it silently re-types an unrelated method with no error at all.

### Cause

For `private_class_method def self.bar`, the `def` node's parent in the nesting
is the modifier call's arguments rather than the enclosing class/module body.
`attached_sigs` walks the `def`'s siblings, so it never sees the `sig` (and
comments) that are declared as siblings of the modifier *call*.

### Fix

When removing a `def` that is wrapped in a modifier call, target the (outermost)
wrapping call instead of the `def`, so its attached sigs and comments are removed
together with the method.

The wrapping call is detected **structurally** (a call that takes the method as an
argument, climbing through nested modifiers) rather than by matching a fixed list of
names. This way user-defined modifiers and future ones — e.g. Sorbet's proposed
[`abstract def`](https://github.com/sorbet/sorbet/pull/10276) — are handled without
maintaining an allowlist.

To stay safe, a call only counts as a modifier when the `def` is its **sole**
argument (and it takes no block). Such a call exists only to wrap that one method,
so removing it whole is correct regardless of the modifier's name. A call that takes
other arguments — e.g. `register(:thing, def bar; end)` or `memoize def bar; end, ttl: 60`
— has its own purpose, so we leave the call (and its sibling arguments) in place and
remove only the `def`. The only remaining ambiguous case, a single-argument
user-defined macro with a side effect, is structurally indistinguishable from a real
modifier and vanishingly rare, so we treat it like one.

After the fix the example above becomes:

```ruby
class Foo
  def baz; end
end
```

### Tests

Added tests covering:

- a visibility modifier (`private_class_method def self.bar`), an arbitrary custom
  modifier, and nested modifiers (`private abstract def bar`) — the sole-argument
  cases where the whole wrapping call (and its sigs/comments) is removed;
- a modifier-wrapped def as the last statement in the body;
- single-line and sig-less modifier forms;
- a `def` passed as a non-sole argument (`register(:thing, def bar; end)`) and a call
  with a trailing keyword argument (`memoize(def bar…, ttl: 60)`) — confirming the
  call and its sibling arguments are preserved and only the `def` is removed.

The full remover suite and `srb tc`/RuboCop pass.

### DIFF
--- a/lib/spoom/deadcode/remover.rb
+++ b/lib/spoom/deadcode/remover.rb
@@ -67,20 +67,15 @@ module Spoom
 
         private
 
-        # When a method is defined as the argument of a modifier call (e.g. `private def foo; end`,
-        # `private_class_method def self.foo; end`, or `abstract def foo; end`), the `def` node's
-        # parent is the modifier call rather than the enclosing class or module body. The sigs and
-        # comments attached to the method are siblings of the outermost such call, so return a
-        # context targeting that call to remove them together with the method.
+        # When a method is defined as the argument of a modifier call (e.g. `private def foo; end` or
+        # `private_class_method def self.foo; end`), the `def` node's parent is the call rather than the
+        # enclosing class or module body, so its sigs and comments are siblings of the call. Return a
+        # context targeting the outermost such call so they are removed together with the method.
         #
-        # Any modifier is matched structurally (rather than from a fixed list) so user-defined and
-        # future modifiers are handled too. To stay safe, a call only counts as a modifier when the
-        # wrapped node is its *sole* argument: such a call exists only to wrap that one method, so
-        # removing it whole is correct regardless of the modifier's name. A call that takes other
-        # arguments (e.g. `register(:thing, def foo; end)`) has its own purpose, so we leave it in
-        # place and remove only the `def`. The remaining ambiguous case — a single-argument
-        # user-defined macro with a side effect — is structurally indistinguishable from a real
-        # modifier and vanishingly rare, so we treat it like one.
+        # Modifiers are matched structurally rather than from a fixed list, so user-defined ones are
+        # handled too. A call only counts as a modifier when the `def` is its sole argument and it takes
+        # no block, so we never remove a call that does more than wrap the method (e.g.
+        # `register(:thing, def foo; end)`).
         #: (Prism::DefNode def_node) -> NodeContext?
         def modifier_call_context(def_node)
           wrapped = def_node #: Prism::Node
@@ -91,11 +86,11 @@ module Spoom
           while (ancestor = nesting.pop)
             case ancestor
             when Prism::ArgumentsNode
-              # An arguments node sits between a call and its arguments, keep climbing.
+              # An arguments node sits between a call and its arguments, keep climbing
               next
             when Prism::CallNode
-              # Stop unless this call wraps the node as its sole argument (and takes no block), so we
-              # never delete sibling arguments or a call doing more than wrapping the method.
+              # Stop unless the call wraps the node as its sole argument and takes no block, otherwise
+              # it does more than wrap the method and we must not remove it
               args = ancestor.arguments&.arguments
               break if ancestor.block
               break unless args && args.size == 1 && args.first.equal?(wrapped)
--- a/lib/spoom/deadcode/remover.rb
+++ b/lib/spoom/deadcode/remover.rb
@@ -74,8 +74,8 @@ module Spoom
         #
         # Modifiers are matched structurally rather than from a fixed list, so user-defined ones are
         # handled too. A call only counts as a modifier when the `def` is its sole argument and it takes
-        # no block, so we never remove a call that does more than wrap the method (e.g.
-        # `register(:thing, def foo; end)`).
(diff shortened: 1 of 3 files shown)


lib/spoom/deadcode/remover.rb
```diff
@@ -65,6 +65,54 @@ def apply_edit
...
+        # wrapped node is its *sole* argument: such a call exists only to wrap that one method, so
+        # removing it whole is correct regardless of the modifier's name. A call that takes other
+        # arguments (e.g. `register(:thing, def foo; end)`) has its own purpose, so we leave it in
+        # place and remove only the `def`. The remaining ambiguous case — a single-argument
+        # user-defined macro with a side effect — is structurally indistinguishable from a real
+        # modifier and vanishingly rare, so we treat it like one.
+        #: -> NodeContext?
+        def modifier_call_context
+          wrapped = @node_context.node #: Prism::Node
+          return unless wrapped.is_a?(Prism::DefNode)
```
KaanOzkan: I think we should call this method only for `DefNode` instead of returning early, being more descriptive in the call site will be helpful.

lib/spoom/deadcode/remover.rb
```diff
@@ -65,6 +65,54 @@ def apply_edit
...
+
+          nesting = @node_context.nesting
+          call_index = nil #: Integer?
+
+          index = nesting.size - 1
```
KaanOzkan: What do you think about iterating the nesting directly? Feels easier to read, something like

```rb
#: (Prism::Node) -> NodeContext?
def modifier_call_context(wrapped)
  nesting = @node_context.nesting.dup
  outer_call = nil #: Prism::CallNode?
  outer_nesting = nil #: Array[Prism::Node]?

  while (ancestor = nesting.pop)
    case ancestor
    when Prism::ArgumentsNode
      next
    when Prism::CallNode
      args = ancestor.arguments&.arguments
      break if ancestor.block
      break unless args && args.size == 1 && args.first.equal?(wrapped)

      wrapped = ancestor
      outer_call = ancestor
      outer_nesting = nesting.dup
    else
      break
    end
  end

  return unless outer_call && outer_nesting

  NodeContext.new(@old_source, @node_context.comments, outer_call, outer_nesting)
end
```

lib/spoom/deadcode/remover.rb
dduugg: Done — split the `case` in `apply_edit` so `Class`/`ModuleNode` delete directly and `modifier_call_context` is only called for `DefNode`, which it now takes as a typed parameter. Dropped the internal early-return guard.

lib/spoom/deadcode/remover.rb
dduugg: Done, adopted the pop-based iteration — reads much better, thanks. One small tweak vs. your snippet: `wrapped` is held in a `Prism::Node` local since it starts as the `DefNode` param and gets reassigned to the wrapping `CallNode`, which would otherwise be a Sorbet type error.

rbi/spoom.rbi
```diff
@@ -1519,6 +1519,9 @@ class Spoom::Deadcode::Remover::NodeRemover
   end
   def insert_accessor(node, send_context, was_removed:); end
 
+  sig { returns(T.nilable(::Spoom::Deadcode::Remover::NodeContext)) }
+  def modifier_call_context; end
```
KaanOzkan: This declaration is now outdated

dduugg: Good catch — regenerated `rbi/spoom.rbi` with `bin/export-rbi` so `modifier_call_context` now reflects the `def_node` parameter.

lib/spoom/deadcode/remover.rb
```diff
@@ -103,6 +103,7 @@ def modifier_call_context(def_node)
             end
           end
           return unless outer_call && outer_nesting
+          return unless outer_nesting.last.is_a?(Prism::StatementsNode)
```
KaanOzkan, line 106: Added this check to cover an edge case LLM pointed out.

KaanOzkan (approved):

dduugg:
TYSM for getting this over the line, @KaanOzkan ! 💜

PR #943, opened 2026-06-08
Preserve heredoc body when translating multi-line `T.let` assertions

styrmis:
When `T.let(...)` spans multiple lines and wraps a heredoc, the replacement range covers the entire call including the heredoc body. Since the value node's source range only covers the opener line (e.g. `<<~MSG.strip`), the body and terminator were silently dropped, producing syntactically broken Ruby.

### DIFF
--- a/lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
+++ b/lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
@@ -250,6 +250,17 @@ module Spoom
           offsets
         end
 
+        #: (Prism::Node) -> bool
+        def string_literal?(node)
+          case node
+          when Prism::StringNode, Prism::InterpolatedStringNode,
+               Prism::XStringNode, Prism::InterpolatedXStringNode
+            true
+          else
+            false
+          end
+        end
+
         #: (Prism::Node, Prism::Node) -> String
         def dedent_value(assign, value)
           if value.location.start_line == assign.location.start_line
@@ -281,7 +292,7 @@ module Spoom
           # ```
           indent = value.location.start_line - assign.location.start_line
           lines = value.slice.lines
-          if lines.size > 1
+          if lines.size > 1 && !string_literal?(value)
             lines[1..]&.each_with_index do |line, i|
               lines[i + 1] = line.delete_prefix("  " * indent)
             end
--- a/lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
+++ b/lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
@@ -103,10 +103,12 @@ module Spoom
           # Otherwise, replace up to the end of the node
           end_offset = comment_end_offset || node.location.end_offset
 
+          heredoc_body = heredoc_body_within_range(value, end_offset)
+
           replacement = if node.name == :bind
             "#{rbs_annotation}#{trailing_comment}"
           else
-            "#{dedent_value(node, value)} #{rbs_annotation}#{trailing_comment}"
+            "#{dedent_value(node, value)} #{rbs_annotation}#{trailing_comment}#{heredoc_body}"
           end
 
           @rewriter << Source::Replace.new(start_offset, end_offset - 1, replacement)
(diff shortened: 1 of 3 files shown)


lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
```diff
@@ -212,6 +214,47 @@ def extract_trailing_comment(node)
...
 
+        #: (Prism::Node, Integer) -> String?
+        def heredoc_body_within_range(node, replace_end_offset)
+          heredoc_end = find_heredoc_end_offset(node)
+          return unless heredoc_end
+          return if heredoc_end > replace_end_offset
+
+          value_end = node.location.end_offset
+          opener_line_end = value_end
+          opener_line_end += 1 while opener_line_end < @ruby_bytes.size && @ruby_bytes[opener_line_end] != LINE_BREAK
```
KaanOzkan: You can use https://github.com/Shopify/spoom/blob/a2a57127c6337e368fbc6272dc7a2bbf1dcdf5a7/lib/spoom/sorbet/translate/translator.rb#L55

lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
```diff
@@ -212,6 +214,47 @@ def extract_trailing_comment(node)
...
+          return if opener_line_end >= @ruby_bytes.size
+
+          body_bytes = @ruby_bytes[(opener_line_end + 1)...heredoc_end] #: as !nil
+          body = body_bytes.pack("C*")
+          body.chomp! if @ruby_bytes[replace_end_offset] == LINE_BREAK
+          "\n#{body}"
+        end
+
+        #: (Prism::Node) -> Integer?
+        def find_heredoc_end_offset(node)
```
KaanOzkan: If we have something like 
```rb
both = T.let(
  foo(<<~A, <<~B),
    first
  A
    second
  B
  String,
)
```
stopping at the first heredoc would end up dropping subsequent ones. I think we can support this case too.

Something like

```rb
#: (Prism::Node) -> Array[Integer]
def heredoc_end_offsets(node)
  offsets = [] #: Array[Integer]

  case node
  when Prism::StringNode, Prism::InterpolatedStringNode
    opening = node.opening_loc
    closing = node.closing_loc

    if opening && closing && opening.start_line != closing.start_line
      offsets << closing.end_offset
    end
  end

  node.each_child_node do |child|
    offsets.concat(heredoc_end_offsets(child))
  end

  offsets
end
```

Using `each_child_node` should take care of it for us. And then inside `heredoc_body_within_range` we can do

```rb
heredoc_end = heredoc_end_offsets(node)
  .select { |offset| offset <= replace_end_offset }
  .max
return unless heredoc_end
```

styrmis: Interesting, I wasn't aware that heredocs could be combined in this manner! I have made both suggested changes.

lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
```diff
@@ -212,6 +214,42 @@ def extract_trailing_comment(node)
...
+
+        #: (Prism::Node) -> Array[Integer]
+        def heredoc_end_offsets(node)
+          offsets = [] #: Array[Integer]
+
+          case node
+          when Prism::StringNode, Prism::InterpolatedStringNode
+            opening = node.opening_loc
+            closing = node.closing_loc
+            if opening && closing && opening.start_line != closing.start_line
```
KaanOzkan: I guess you could split a string in multiple lines 😄 
```rb
s = T.let(
  "first
  second",
  String,
)
```

I found [slice](https://github.com/ruby/prism/blob/v1.9.0/lib/prism/parse_result.rb#L363-L365), so we can add a `opening.slice.start_with?("<<")` check to the conditional:
`opening && closing && opening.start_line != closing.start_line && opening.slice.start_with?("<<")`

styrmis: Nice catch—have separately added coverage for backticks in heredocs.

test/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments_test.rb
```diff
@@ -264,6 +264,93 @@ def test_translate_assigns_ignore_heredoc_values
...
+            s = T.let(
+              "first
+              second",
+              String,
+            )
+          RB
+
+          assert_equal(<<~RB, rbi_to_rbs(rb))
+            s = "first
+            second" #: String
```
KaanOzkan:
```suggestion
              second" #: String
```
I think preserving the indentation is better than not preserving it, wdyt?

styrmis: Completely agree—looks like this is existing behaviour of `dedent_value` but the fix is small and it doesn't affect any other tests so have applied it in 5b2aa5169389b15a7e2ed8f976e5498f004c77df 👍

KaanOzkan (approved):
Thanks

PR #948, opened 2026-06-09
Skip RBS type aliases as node signatures

### LINKED ISSUE #947: RBS type alias above method sig results in incorrect translation
```rb
require "spoom/sorbet/translate"

src = <<~RUBY
  # typed: strict
  class X
    #: type foo = Integer
    #: (foo) -> void
    def m(x); end
  end
RUBY

puts Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs(src, file: "x.rb")
```

outputs

```rb
# typed: strict
class X
  Foo = T.type_alias { Integer }
d
end
```

Note the syntax error second to last line and that the method definition is gone.

KaanOzkan:
Fixes https://github.com/Shopify/spoom/issues/947

Consumers of `node_rbs_comments` shouldn't receive a type alias comment. It's handled seperately in https://github.com/Shopify/spoom/blob/a2a57127c6337e368fbc6272dc7a2bbf1dcdf5a7/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb#L368

### DIFF
--- a/lib/spoom/rbs.rb
+++ b/lib/spoom/rbs.rb
@@ -105,6 +105,9 @@ module Spoom
               location = location.join(continuation_comment.location)
             end
             continuation_comments.clear
+
+            next if string.start_with?("type ")
+
             res.signatures.prepend(Signature.new(string, location))
           elsif string.start_with?("#|")
             continuation_comments << comment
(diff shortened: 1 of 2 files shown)


paracycle (approved):
The approach is a little ad-hoc, but I think that's fine.

amomchilov (approved):

PR #949, opened 2026-06-12
Rename `RBSCommentsToSorbetSigs` to `HumanReadableTranslator`

amomchilov:
## Breaking change

This PR makes a breaking change of renaming the `RBSCommentsToSorbetSigs` class to `RBSCommentsToSorbetSigs::HumanReadableTranslator`

These two APIs will still work as before:

* `RBSCommentsToSorbetSigs.contains_rbs_syntax?`
* `RBSCommentsToSorbetSigs.rewrite_if_needed`

But you can no longer do `RBSCommentsToSorbetSigs.new`. Instead, you must pick a particular translation strategy. There's currently just one: `RBSCommentsToSorbetSigs::HumanReadableTranslator.new`.

## Why?

For #787 wrote a second strategy called `RBSCommentsToSorbetSigs::LineMatchingTranslator`, and it's easier to pick the class than to make a complicated wrapper that tries to pick for you.

# How

This PR does this refactor in 3 green steps, which preserve the git history intact.

Review the 3 commits separately, and the diff is minimal.

### DIFF
--- a/lib/spoom/sorbet/translate.rb
+++ b/lib/spoom/sorbet/translate.rb
@@ -5,7 +5,7 @@ require "rbi"
 
 require "spoom/source/rewriter"
 require "spoom/sorbet/translate/translator"
-require "spoom/sorbet/translate/rbs_comments_to_sorbet_sigs"
+require "spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/human_readable_translator"
 require "spoom/sorbet/translate/sorbet_assertions_to_rbs_comments"
 require "spoom/sorbet/translate/sorbet_sigs_to_rbs_comments"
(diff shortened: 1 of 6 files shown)


KaanOzkan (approved):
Regarding the API I thought about defining a `new` on `RBSCommentsToSorbetSigs` that defaults to `HumanReadableTranslator` but I don't think it's necessary, these should be used internally in Spoom so it makes sense to be explicit.

amomchilov:
@KaanOzkan I initially had that, but it got really gnarly once I needed to introduce an `options:` kwarg to the parameter list. I made a new Options class to hold the two current kwargs (plus some new options), but to stay backwards compatible I'd need to accept either and handle them both, which was finicky

I gave up on it and decided to just do a clean break

PR #946, opened 2026-06-08
Add rewriter option to erase generics

dejmedus:
- part of https://github.com/Shopify/spoom/issues/924

> May be easier to review by commit :)

Sorbet ignores generic types at runtime as they [cannot be enforced](https://sorbet.org/docs/stdlib-generics#generics-and-runtime-checks). This means in runtime contexts we can drop generic types entirely, simplifying the RBS to Sorbet signature translation

This PR adds an `erase_generic_types` option to the `RBSCommentsToSorbetSigs` rewriter and `--erase-generic-types flag` to CLI. When enabled generic `Foo[Bar]` will become `Foo` and type members will become type aliases of `T.anything`

PR #953, opened 2026-06-16
Use a TypeTranslator instance in RBSCommentsToSorbetSigs

### LINKED ISSUE #924: Add rewriter option for erasing generics
## Motivation

Sorbet runtime can't type-check generics at runtime:

<details><summary>Example error not being detected at runtime</summary>

```ruby
#!/usr/bin/ruby

require "sorbet-runtime"

class Box
  extend T::Generic
  
  Value = type_member
  
  def initialize(value)
    @value = value
  end
end

extend T::Sig

sig { returns(Box[Integer]) }
def box_of_int = Box[String].new("Not an integer!")

box_of_int() # no runtime error.
```

</details>

If we're using the Spoom rewriter for the sole purpose of reifying RBS sigs to enable type-checking in CI, then there's no benefit for us to keep the generic types around.

E.g. there's no reason to `extend T::Generic`, if we don't need the `[]` method.

## Proposed solution

Add a new setting like `erase_generic_types: bool` to `RBSCommentsToSorbetSigs`.

For this example input:

```ruby
#: [E]
class Box
  #: -> void
  def initialize
    @elems = [] #: T::Array[E]
  end
end

#: -> Box[Integer]
def box_of_int = Box.new(123) #: Box[Integer]
```

When enabled, the rewritten output would erase generics, like so:

```diff
 class Box
- extend T::Generic
- 
- E = type_member
  
  sig { void }
  def initialize
    @elems = [] #: T::Array[E]
  end
 end
 
-sig { returns(Box[Integer]) }
+sig { returns(Box) }
 def box_of_int = Box.new(123) #: Box[Integer]
```

This would also involved changing how method signatures are translated, so we'd need to add  them same setting to [`RBI::RBS::MethodTypeTranslator`](https://github.com/Shopify/rbi/blob/main/lib/rbi/rbs/method_type_translator.rb) in the `rbi` gem first.

dejmedus:
- Part of https://github.com/Shopify/spoom/issues/924

Previously, `RBSCommentsToSorbetSigs` translated types using the `RBI::RBS::TypeTranslator` `translate` method. In order to support the ability to pass translation configuration. This PR creates a `TypeTranslator` instance and updates type translation callers to use it

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
@@ -22,6 +22,7 @@ module Spoom
 
             @max_line_length = max_line_length
             @overloads_strategy = overloads_strategy
+            @type_translator = RBI::RBS::TypeTranslator.new #: RBI::RBS::TypeTranslator
           end
 
           # @override
@@ -108,11 +109,11 @@ module Spoom
                 name = node.arguments&.arguments&.first #: as Prism::SymbolNode
                 sig.params << RBI::SigParam.new(
                   name.slice[1..-1], #: as String
-                  RBI::RBS::TypeTranslator.translate(attr_type),
+                  @type_translator.translate(attr_type),
                 )
               end
 
-              sig.return_type = RBI::RBS::TypeTranslator.translate(attr_type)
+              sig.return_type = @type_translator.translate(attr_type)
 
               apply_member_annotations(comments.method_annotations, sig)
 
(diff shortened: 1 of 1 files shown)


dejmedus:
* **#946** <a href="https://app.graphite.com/github/pr/Shopify/spoom/946?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#953** <a href="https://app.graphite.com/github/pr/Shopify/spoom/953?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Shopify/spoom/953?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

amomchilov (approved):

PR #965, opened 2026-06-23
Allow skipping abstract method translation

vinistock:
This PR introduces a new option that allows skipping the translation of abstract methods.

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
@@ -25,6 +25,7 @@ module Spoom
             end #: Integer?
 
             @overloads_strategy = options.overloads_strategy #: Symbol
+            @translate_abstract_methods = options.translate_abstract_methods #: bool
             @type_translator = RBI::RBS::TypeTranslator.new #: RBI::RBS::TypeTranslator
           end
 
@@ -141,6 +142,7 @@ module Spoom
           def rewrite_def(def_node, comments)
             return if comments.empty?
             return if comments.signatures.empty?
+            return if !@translate_abstract_methods && comments.method_annotations.any?(&:abstract?)
 
             signatures = apply_overloads_strategy(
               comments.signatures,
--- a/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
+++ b/test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
@@ -147,6 +147,66 @@ module Spoom
           )
         end
 
+        def test_translate_to_rbi_skipping_abstract_methods
+          assert_rewrites_rbs(
+            from: <<~RUBY,
+              # @abstract
+              class Foo
+                # @abstract
+                #: -> String
+                def bar; end
+              end
+
+              # @abstract
+              module Baz
+                # @abstract
+                #: -> String
+                def qux; end
+              end
+            RUBY
+
+            to_pretty_format_for_humans: <<~RUBY,
+              class Foo
+                extend T::Helpers
+
+                abstract!
+
+                # @abstract
+                #: -> String
+                def bar; end
+              end
+
+              module Baz
+                extend T::Helpers
+
+                abstract!
+
+                # @abstract
+                #: -> String
+                def qux; end
+              end
+            RUBY
+
+            to_line_matched_format_for_machines: <<~RUBY,
+              # RBS_REWRITTEN_ANNOTATION: @abstract
+              class Foo; extend T::Helpers; abstract!
+                # @abstract
+                #: -> String
+                def bar; end
+              end
+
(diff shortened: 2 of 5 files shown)


lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -141,6 +142,7 @@ def visit_attr(node)
           def rewrite_def(def_node, comments)
             return if comments.empty?
             return if comments.signatures.empty?
+            return if !@translate_abstract_methods && abstract_method?(comments)
```
amomchilov: I think you should add this to `Annotation`:

```diff
class Annotation < Comment
+ #: -> bool
+ def abstract? = "@abstract" == string
end
```

Then you can simplify this to not need a helper:


```suggestion
            return if !@translate_abstract_methods && comments.method_annotations.any?(&:abstract?)
```

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -147,6 +147,68 @@ def bar; end
           )
         end
 
+        def test_translate_to_rbi_skipping_abstract_methods
+          source = <<~RUBY
```
amomchilov: Is there any reason why the `assert_rewrites_rbs` method couldn't be used here?

It does validation to make sure the expected value is correctly preserves line numbering

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -147,6 +147,68 @@ def bar; end
...
+                #: -> String
+                def qux; end
+              end
+            RUBY
+            rbs_comments_to_sorbet_sigs(source, translate_abstract_methods: false),
+          )
+
+          assert_equal(
+            <<~RUBY,
+              # RBS_REWRITTEN_ANNOTATION: @abstract
```
amomchilov, line 191: It's not rewritten.

The other such comment is `RBS_IGNORED_UNKNOWN_ANNOTATION`, which isn't right here (this is known, but intentionally ignore).

Perhaps


```suggestion
              # RBS_IGNORED_ANNOTATION: @abstract
```

vinistock: That's the `@abstract` on the class/module though. The one on methods don't get any prefix at all.

Do you think we should still add one?

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
vinistock: Not really. Switched to using `assert_rewrite_rbs`.

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
amomchilov: Oh I meant to comment this on the method-level annotation.

I didn't implement the `RBS_REWRITTEN_ANNOTATION` comment to method-level annotations, that's just an oversight on my part. Could be nice to add it (not in this PR, and not urgent).

Those comments only exist for debuggability, to help with investigating any potential issues with the output.

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -1280,33 +1340,57 @@ def foo; end
...
+          RBSCommentsToSorbetSigs::LineMatchingTranslator.new(
+            ruby_contents,
+            file: "test.rb",
+            options: RBSCommentsToSorbetSigs::Options.new(
+              translate_abstract_methods:,
+              output_format: RBSCommentsToSorbetSigs::LineMatchedRBIFormat.default,
+            ),
+          ).rewrite
+        end
+
```
amomchilov: this is unused now

```suggestion
```

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -191,6 +193,11 @@ def rewrite_def(def_node, comments)
             end
           end
 
+          #: (Spoom::RBS::Comments) -> bool
+          def abstract_method?(comments)
+            comments.method_annotations.any?(&:abstract?)
```
amomchilov: Eh personally I think it's more clear when inlined, but that's a matter of personal preference

amomchilov (approved):
Thank you!

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
vinistock: I inlined the method 👍

PR #966, opened 2026-06-29
Use typed override for spoom bump

Morriar:
Run Sorbet with a temporary `--typed-override` file when checking bump candidates instead of rewriting sigils before type checking.

### DIFF
--- a/lib/spoom/cli/srb/bump.rb
+++ b/lib/spoom/cli/srb/bump.rb
@@ -3,6 +3,8 @@
 
 require "find"
 require "open3"
+require "tempfile"
+require "yaml"
 
 module Spoom
   module Cli
@@ -72,6 +74,11 @@ module Spoom
             exit(1)
           end
 
+          unless context.sorbet_config.typed_overrides.empty?
+            say_error("Cannot run `spoom bump` on a project that already uses Sorbet's `--typed-override` option")
+            exit(1)
+          end
+
           say("Checking files...")
 
           files_to_bump = context.srb_files_with_strictness(from, include_rbis: false)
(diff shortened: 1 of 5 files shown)


lib/spoom/cli/srb/bump.rb
```diff
@@ -189,8 +193,15 @@ def print_changes(files, command:, from: "false", to: "true", dry: false, path:
             end
           end
 
-          def undo_changes(files, from_strictness)
-            Sorbet::Sigils.change_sigil_in_files(files, from_strictness)
+          #: (Spoom::Context context, Array[String] files, String strictness) -> String
+          def write_typed_override_file(context, files, strictness)
+            path = ".spoom-typed-override-#{Process.pid}-#{SecureRandom.hex(8)}.yaml"
```
amomchilov: Can't we just use `Tempfile.new(['spoom-typed-override', '.yml'])`?

Then there's no need to have it hidden with the prefixed `.`

lib/spoom/cli/srb/bump.rb
```diff
@@ -189,8 +193,15 @@ def print_changes(files, command:, from: "false", to: "true", dry: false, path:
...
+            relative_paths = files.map do |file|
+              relative_path = file.delete_prefix("#{context.absolute_path}/")
+              "./#{relative_path}"
+            end
+            context.write!(path, YAML.dump({ strictness => relative_paths }))
```
juharris: Do we delete the file when we're done?
Is that was using `context` helps with?

amomchilov (approved):

PR #970, opened 2026-07-07
Use correct type for `RBI::Sig#loc` during translation

Morriar:
Found with https://github.com/Shopify/tapioca/pull/2667.


vinistock (approved):
Why didn't Sorbet catch this? Are we missing some signatures somewhere?

Or do we accept both `RBI::Loc` or `Prism::Location`?

Morriar:
> Why didn't Sorbet catch this? Are we missing some signatures somewhere?
> 
> Or do we accept both `RBI::Loc` or `Prism::Location`?

Because we were not generating typed writers for `attr_accessor`, see https://github.com/Shopify/tapioca/pull/2667.

PR #971, opened 2026-07-08
Preserve Sorbet warnings in formatted srb output

KaanOzkan:
Capture leading Sorbet stderr lines while parsing typecheck output so formatted `spoom srb tc` and `spoom srb bump` output no longer drops warnings emitted before errors.

I initially changed the existing `parse_string` function but it's an API used by Tapioca, opted for a new function instead.

### DIFF
--- a/lib/spoom/sorbet/errors.rb
+++ b/lib/spoom/sorbet/errors.rb
@@ -56,22 +56,35 @@ module Spoom
         ] #: Array[String]
 
         class << self
+          # Used when only Sorbet errors are needed and leading stderr warnings can be ignored.
           #: (String output, ?error_url_base: String) -> Array[Error]
           def parse_string(output, error_url_base: DEFAULT_ERROR_URL_BASE)
+            parse_result(output, error_url_base: error_url_base).errors
+          end
+
+          # Used when callers need both parsed Sorbet errors and leading stderr warnings.
+          #: (String output, ?error_url_base: String) -> ParseResult
+          def parse_result(output, error_url_base: DEFAULT_ERROR_URL_BASE)
             parser = Spoom::Sorbet::Errors::Parser.new(error_url_base: error_url_base)
-            parser.parse(output)
+            parser.parse_result(output)
           end
         end
 
         #: (?error_url_base: String) -> void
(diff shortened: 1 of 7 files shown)


lib/spoom/sorbet/errors.rb
```diff
@@ -56,22 +56,35 @@ class ParseError < Spoom::Error; end
         ] #: Array[String]
 
         class << self
+          # Used when only Sorbet errors are needed and leading stderr warnings can be ignored.
           #: (String output, ?error_url_base: String) -> Array[Error]
           def parse_string(output, error_url_base: DEFAULT_ERROR_URL_BASE)
+            parse_result(output, error_url_base: error_url_base).errors
+          end
```
Morriar, line 63: Can't we just change this one to return a `ParseResult`?

lib/spoom/sorbet/errors.rb
```diff
@@ -56,22 +56,35 @@ class ParseError < Spoom::Error; end
...
+          @warnings = [] #: Array[String]
           @error_line_match_regex = error_line_match_regexp(error_url_base) #: Regexp
           @current_error = nil #: Error?
         end
 
         #: (String output) -> Array[Error]
         def parse(output)
+          parse_result(output).errors
+        end
+
```
Morriar, line 85: Same here, let's just return a ParseResult directly?

lib/spoom/sorbet/errors.rb
KaanOzkan: That's what I had initially but Tapioca uses `parse_string` in [here](https://github.com/Shopify/tapioca/blob/47f5e8f2fe168f70d5527266073934dd0ab884bb/lib/tapioca/helpers/rbi_files_helper.rb#L87). Which I had to execute during https://github.com/Shopify/spoom/blob/main/bin/export-rbi. So I went with a new function instead to be backwards compatible and ship this easier.

lib/spoom/sorbet/errors.rb
KaanOzkan: We could although it's public API and would break callers if there are any

Morriar (approved):

PR #973, opened 2026-07-13
Support RBS comment translation on multi-name attrs

KaanOzkan:
Will fix this issue after a Spoom bump in tapioca https://github.com/Shopify/tapioca/issues/2662

Rewrite `attr_reader`, `attr_writer`, and `attr_accessor` calls with multiple arguments so that each generated method has its own pending Sorbet `sig`.

As a result of this PR

```ruby
#: Integer
attr_reader :a, :b
```
is translated into
```rb
sig { returns(Integer) }
attr_reader :a

sig { returns(Integer) }
attr_reader :b
```

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
@@ -95,6 +95,11 @@ module Spoom
 
             return if comments.signatures.empty?
 
+            attr_name_nodes = node.arguments&.arguments || []
+            return if attr_name_nodes.empty?
+
+            sig_strings_by_attr_name = attr_name_nodes.map { [] } #: Array[Array[String]]
+
             signatures = apply_overloads_strategy(
               comments.signatures,
               method_name: node.message.to_s,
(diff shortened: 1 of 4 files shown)


Morriar (commented):
I'm guessing this is related to https://github.com/Shopify/tapioca/issues/2662. Shouldn't this be fixed at the `sorbet-runtime` layer instead?

KaanOzkan:
@Morriar I gave it a go in https://github.com/Shopify/sorbet/pull/870/changes but couldn't reduce the surface area. We need to keep track of multiple `method_added` calls and associate it with the same signature. I feel like fixing it in Spoom is a cleaner implementation, wdyt?

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -92,51 +92,154 @@ def visit_call_node(node)
...
+              raise Error, "RBS signatures for attributes do not support overloads"
+            end
 
-            signatures = apply_overloads_strategy(
-              comments.signatures,
-              method_name: node.message.to_s,
-              location: "#{@file}:#{node.location.start_line}",
-            )
+            signature = signatures.first
+            return unless signature
```
soutaro: `raise` instead of `return` or `#: as !nil` look better to me, since this is unreachable.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -92,51 +92,154 @@ def visit_call_node(node)
           #: (Prism::CallNode) -> void
           def visit_attr(node)
             comments = node_rbs_comments(node)
-            return if comments.empty?
+            signatures = comments.signatures
+            return if signatures.empty?
 
-            return if comments.signatures.empty?
+            if signatures.size != 1
```
soutaro: Is it intentional that this ignores `overloads_strategy` and always raises? I think this is reasonable, but I wanted to confirm.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -92,51 +92,154 @@ def visit_call_node(node)
...
+            signature = signatures.first
+            return unless signature
 
-            known_annotations = nil #: Array[Spoom::RBS::Annotation]?
+            attr_name_nodes = node.arguments&.arguments || []
 
-            signatures.each do |signature|
-              attr_type = ::RBS::Parser.parse_type(signature.string)
-              sig = RBI::Sig.new
+            if node.message == "attr_writer" && attr_name_nodes.size != 1
```
soutaro: Now that the code generates a separate `sig` for each attribute, it seems possible to support `attr_writer` with multiple names by removing this validation.

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -317,24 +317,196 @@ class A
...
+          )
+        end
+
+        def test_translate_to_rbi_preserves_trailing_comments_in_multi_name_attrs
+          assert_rewrites_rbs(
+            from: <<~RUBY,
+              class A
+                #: Integer
+                attr_reader(
+                  # first attribute
```
soutaro, line 484: It would be great if these comments could be preserved somewhere in the rewritten source so users can decide how to handle them.

soutaro (changes requested):

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
KaanOzkan: It's not supported by Sorbet so I explicitly errored in this case [sorbet.run](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20true%0A%0Aclass%20Current%0A%20%20%23%3A%20String%0A%20%20%23%3A%20Integer%0A%20%20attr_reader%20%3Afoo%0Aend)

KaanOzkan: I guess if it's not a lot of work we should support it in the translation and let Sorbet error, I'll change it.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
KaanOzkan: I was too focused on https://github.com/Shopify/spoom/pull/973#discussion_r3925410289, bringing `overloads_strategy` back, thanks.

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
soutaro: I was assuming the code was to reject `attr_writer :foo, :bar`. But it looks good now. 👍

soutaro (approved):

PR #977, opened 2026-07-15
Fix RBS rewriter for methods with anonymous keyword/positional rest params

rafaelfranca:
When a method definition uses anonymous rest params (e.g. `def resolve(**)` or `def resolve(*)`), the RBS-to-Sorbet sig translation produced invalid output: `sig { params(**kwargs: ::T.untyped)... }` -- a SyntaxError that
prevented the entire file from loading.

Root cause: RBI::RBS::MethodTypeTranslator falls back to the RBI parser's synthetic names (`**kwargs`, `*args`) for anonymous rest params. These splat-prefixed names are invalid in Sorbet sig params.

Fix: When the method def has anonymous `**` or `*` (detected via Prism), the rewriter normalizes the corresponding translated SigParam names to quoted star names (`"**"` / `"*"`) so the RBI printer emits
`params("**": ...)` / `params("*": ...)`.

sorbet-runtime accepts these quoted star names as sig param names for anonymous rest params, so the sig binds correctly without rewriting the method def.  This means anonymous forwarding expressions in the body
(e.g. `g(**)` or `g(*)`) remain valid Ruby -- no body rewriting needed.

RBS param names are non-semantic, so even when the RBS type names the param (e.g. `(**untyped kwargs)`), the sig uses the quoted star name to match the anonymous method def param.

The normalization is applied per translated signature, so malformed RBS leaves the method untouched (no sig emitted, no changes).

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
@@ -144,15 +144,22 @@ module Spoom
             return if comments.signatures.empty?
             return if !@translate_abstract_methods && comments.method_annotations.any?(&:abstract?)
 
+            builder = RBI::Parser::TreeBuilder.new(@ruby_contents, comments: [], file: @file)
+            builder.visit(def_node)
+            rbi_node = builder.tree.nodes.first #: as RBI::Method
+
+            # When the method def has anonymous rest/keyword-rest params
+            # (e.g. `def foo(**)` or `def foo(*)`), sorbet-runtime cannot
+            # bind a named sig param to them.  Instead of rewriting the
+            # method def, we emit the sig with quoted star names (`"*"` /
+            # `"**"`) which sorbet-runtime accepts for anonymous rest params.
+            anonymous_overrides = find_anonymous_rest_overrides(def_node, rbi_node)
+
             signatures = apply_overloads_strategy(
               comments.signatures,
               method_name: def_node.name.to_s,
               location: "#{@file}:#{def_node.location.start_line}",
             )
-
-            builder = RBI::Parser::TreeBuilder.new(@ruby_contents, comments: [], file: @file)
-            builder.visit(def_node)
-            rbi_node = builder.tree.nodes.first #: as RBI::Method
 
             known_annotations = nil #: Array[Spoom::RBS::Annotation]?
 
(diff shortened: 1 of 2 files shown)


lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -193,6 +263,90 @@ def rewrite_def(def_node, comments)
...
+            rewrites = [] #: Array[Source::Replace]
+
+            # Anonymous ** — name is deliberately ugly to avoid collisions.
+            kr = prism_params.keyword_rest
+            if kr.is_a?(Prism::KeywordRestParameterNode) && kr.name.nil?
+              name = "__sorbet_rbs_kwargs__"
+              rewrites << Source::Replace.new(
+                kr.location.start_offset,
+                kr.location.end_offset - 1,
+                "**#{name}",
```
amomchilov: This changes the actual Ruby code (observable through `method(__method__).parameters`, `binding`, etc.).

The names can stay as `*` and `**`, but they need to be printed with the quoted Symbol syntax ([Sorbet.run](https://sorbet.run/#%23%20typed%3A%20true%0A%0Aextend%20T%3A%3ASig%0A%0Asig%20%7B%20params%28%22*%22%3A%20Integer%29.void%20%7D%0Adef%20has_anon_rest%28*%29%3B%20end%0A%0Asig%20%7B%20params%28%22**%22%3A%20Integer%29.void%20%7D%0Adef%20has_anon_kwargs%28**%29%3B%20end%0A%0A%23%20This%20one%20is%20not%20supported%20yet.%0Asig%20%7B%20params%28%22%26%22%3A%20Integer%29.void%20%7D%0Adef%20has_anon_block%28%26%29%3B%20end)):

```ruby
# typed: true

extend T::Sig

sig { params("*": Integer).void }
def has_anon_rest(*); end

sig { params("**": Integer).void }
def has_anon_kwargs(**); end

# This one is not supported yet.
sig { params("&": Integer).void }
def has_anon_block(&); end
```

We won't be able to do the same for `&`, though.

amomchilov (changes requested):

rafaelfranca:
I left `&` out and change the approach to use string as keys

Morriar:
@rafaelfranca  I think https://github.com/Shopify/rbi/pull/624 is a better approach

rafaelfranca:
Yes. I'm fine with that

PR #979, opened 2026-07-17
Fix ERB deadcode parsing for Rails layout yields

### LINKED ISSUE #975: Deadcode ERB parser rejects Rails layout `yield`
Spoom's dead-code analyzer skips valid Rails layout templates containing `<%= yield %>` because the generated Ruby is parsed outside a method context.

## Reproduction

Create `view.erb`:

```erb
<%= yield %>
```

Run:

```sh
bundle exec spoom deadcode deadcode view.erb --no-color
```

With Spoom 1.8.3, the output includes:

```text
Collecting files...
Indexing 1 files...
Error: Error parsing view.erb: Error while parsing view.erb:
- Invalid yield (at 1:24)
Analyzing 0 definitions against 0 references...

No dead code found!
```

I reproduced this with:

- Spoom 1.8.3
- Ruby 4.0.1

## Expected behavior

Spoom parses the valid Rails layout template and indexes any other Ruby references in it.

## Actual behavior

`Spoom::Deadcode::ERB` compiles the expression to Ruby resembling:

```ruby
@output_buffer.append=( yield )
```

Spoom then parses that source at the top level. Ruby only permits `yield` in a method context, so parsing fails. The dead-code command catches the parse error, skips the whole template, and continues.

This is more than a noisy warning: references elsewhere in the skipped layout are omitted from the index and can produce false-positive dead-code candidates.

Rails compiles templates into methods, where this use of `yield` is valid.

## Possible fix

Teach `Spoom::Deadcode::ERB` or `Index#index_erb` to parse Rails layout yields in their method context, or replace yield-only output expressions with a parseable placeholder before indexing. A placeholder is sufficient for a bare `yield` because it does not itself reference a Ruby method or constant, but forms with arguments should preserve indexable references in those arguments.

A regression test could extend the existing ERB tests with a layout containing both a yield and an ordinary reference:

```erb
<%= LayoutHelper.title %>
<%= yield %>
```

The test should assert that parsing succeeds and `LayoutHelper` and `title` are indexed.

## Related work

- [Deadcode: handle ERB files](https://github.com/Shopify/spoom/pull/375)
- [Fix ERB parsing when running deadcode CLI command](https://github.com/Shopify/spoom/pull/555)

Morriar:
## Summary

Fixes #975.

Spoom's deadcode indexer parses ERB templates by first compiling them through our Rails-style Erubi wrapper, then parsing the generated Ruby source.

Rails layout templates can contain bare `<%= yield %>`, which compiles to Ruby using `yield`. That is valid in Rails because templates are compiled into methods, but Spoom was parsing the generated Ruby at the top level, where `yield` is invalid.

This updates ERB indexing to parse generated template code inside a synthetic method, matching Rails' compilation context more closely.

## Changes

* Update the custom Erubi wrapper to use the current Rails-style buffer wrapping behavior.
* Add `wrapped_src` for ERB templates so generated code is parsed inside a synthetic method.
* Ignore the synthetic ERB wrapper method so it does not appear as a deadcode candidate.
* Add a regression test for Rails layout templates containing both a normal reference and `<%= yield %>`.


lib/spoom/deadcode/erb.rb
```diff
@@ -50,32 +59,34 @@ def add_text(text)
...
         else
-          src << bufvar << ".safe_append='"
-          src << "\n" * @newline_pending if @newline_pending > 0
-          src << text.gsub(/['\\]/, '\\\\\&')
-          src << "'.freeze;"
-
+          with_buffer do
+            src << ".safe_append='"
+            src << "\n" * @newline_pending if @newline_pending > 0
+            src << text.gsub(/['\\]/, '\\\\\&') << @text_end
```
amomchilov, line 65: What's this doing?

amomchilov (approved):

lib/spoom/deadcode/erb.rb
Morriar: As mentioned at the top of the file, this code is copied verbatim from https://github.com/rails/rails/blob/main/actionview/lib/action_view/template/handlers/erb/erubi.rb.

I just updated it to keep the same implementation since the upstream changed.

This particular line escapes single quotes and slashes:

```rb
"it's".gsub(/['\\]/, '\\\\\&') # => "it\\'s"
```

and

```ruby
"a\\b".gsub(/['\\]/, '\\\\\&') # => "a\\\\b"
```

to avoid breaking the generated Ruby that uses a single quote string delimiter on `safe_append`:

```ruby
buffer.safe_append='<text>'
```

PR #996, opened 2026-08-05
Skip assigned T.bind assertion translations

Morriar:
## Why?

Given this Sorbet assertion:

```ruby
bound = T.bind(self, FulfillmentOrderMoverTest)
```

`spoom srb assertions translate` currently produces:

```ruby
bound = #: self as FulfillmentOrderMoverTest
```

This is invalid Ruby. Unlike `T.let` and `T.cast`, `T.bind` translates to a standalone RBS comment and leaves no expression that can remain on the right-hand side of the assignment. Until there is a safe translation for that shape, assignment-wrapped `T.bind` calls should be left unchanged.

## What changed?

* Leave assignment-wrapped `T.bind` calls unchanged for now.
* Add regression coverage for the failing assignment case.

### DIFF
--- a/test/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments_test.rb
+++ b/test/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments_test.rb
@@ -595,6 +595,14 @@ module Spoom
           RB
         end
 
+        def test_skip_bind_in_assignment
+          rb = <<~RB
+            bound = T.bind(self, FulfillmentOrderMoverTest)
+          RB
+
+          assert_equal(rb, rbi_to_rbs(rb))
+        end
+
         def test_translate_options
           rb = <<~RB
             T.bind(self, T.class_of(String))
--- a/lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
+++ b/lib/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments.rb
@@ -69,6 +69,11 @@ module Spoom
 
         #: (Prism::Node) -> bool
         def maybe_translate_assertion(node)
+          # Most assertions can replace the value on the right-hand side of an assignment.
+          # `T.bind` becomes a standalone RBS comment instead, so replacing an assigned
+          # `T.bind` would leave invalid code such as `bound = #: self as Type`.
+          standalone_assertion = node.is_a?(Prism::CallNode)
+
           node = case node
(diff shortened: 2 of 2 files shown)


test/spoom/sorbet/translate/sorbet_assertions_to_rbs_comments_test.rb
```diff
@@ -595,6 +595,14 @@ def test_translate_bind
           RB
         end
 
+        def test_skip_bind_in_assignment
+          rb = <<~RB
+            bound = T.bind(self, FulfillmentOrderMoverTest)
```
paracycle, line 600: Wouldn't it be better to rewrite this to:
```
bound = self #: self as FulfillmentOrderMoverTest
```
instead?

Morriar: This is invalid syntax: https://sorbet.run/?arg=--parser=prism&arg=--print=rewrite-tree&arg=--enable-experimental-rbs-comments#%23%20typed%3A%20true%0A%0Afoo%20%3D%20self%20%23%3A%20self%20as%20Integer

The problem here is that the bind applies to both `bound` and `self` which I don't think was expected: https://sorbet.run/?arg=--parser=prism&arg=--print=rewrite-tree&arg=--enable-experimental-rbs-comments#%23%20typed%3A%20true%0A%0Afoo%20%3D%20T.bind%28self%2C%20Integer%29%0AT.reveal_type%28self%29%0AT.reveal_type%28foo%29

I think what they really wanted was a cast: https://sorbet.run/?arg=--parser=prism&arg=--print=rewrite-tree&arg=--enable-experimental-rbs-comments#%23%20typed%3A%20true%0A%0Afoo%20%3D%20T.cast%28self%2C%20Integer%29%0AT.reveal_type%28self%29%0AT.reveal_type%28foo%29

So my plan is to avoid them for now and fix them with a Rubocop.

paracycle: We can rewrite to:
```ruby
#: self as FulfillmentOrderMoverTest
bound = self
```
though, and that seems to work for both requirements: https://sorbet.run/?arg=--parser=prism&arg=--enable-experimental-rbs-comments#%23%20typed%3A%20true%0A%0A%23%3A%20self%20as%20Integer%0Afoo%20%3D%20self%0A%0AT.reveal_type%28foo%29

PR #1000, opened 2026-08-06
Add force option to RBS comment translation

Morriar:
## Why?

`rbs_comments_to_sorbet_sigs` currently skips files without a valid Sorbet sigil, even when they contain RBS comments. Some callers need to translate those files explicitly (see https://github.com/Shopify/rubocop-sorbet/pull/397/changes#diff-6ba746ed3dcb37240ffd44cd598c2d3422c91bdadd54c8d2b8dcaf30c92006f0R188-R201).

## What changed?

Added an optional `force:` argument to `rbs_comments_to_sorbet_sigs`.

When enabled, RBS comments are translated without requiring a valid `# typed:` sigil. The source must still contain supported RBS syntax, and the default behavior remains unchanged.

This option is limited to the Ruby API; CLI behavior is unchanged.


dejmedus (approved):
LGTM:)

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
```diff
@@ -29,15 +29,18 @@ def contains_rbs_syntax?(source)
...
           def rewrite_if_needed(
             ruby_contents,
             file:,
             max_line_length: nil,
             overloads_strategy: :translate_all,
-            erase_generic_types: false
+            erase_generic_types: false,
+            force: false
           )
-            return ruby_contents unless contains_rbs_syntax?(ruby_contents)
```
KaanOzkan, line 40: Should we delete the `contains_rbs_syntax` definition? Not used but public.

KaanOzkan (approved):

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs.rb
Morriar: Good catch, removed 👍

PR #1001, opened 2026-08-06
Fully qualify `extend T::Generic` & `T::Helpers`

### LINKED ISSUE #990: Generic types called `T` can lead to `TypeError`
Unless you specify `erase_generic_types: true`, this code:

```ruby
#: [T]
class Box
  extend T::Helpers

  ...
end
```

rewrites into:

```ruby
#: [T]
class Box
  extend T::Generic
  T = type_member

  extend T::Helpers   # 💥 raises NameError
  
  ...
end
```

Fix: if the rewriter is going to insert an `extend T::Generic`, it should be a fully-qualified `extend ::T::Generic`. Likewise for `extend ::T::Helpers`.

dejmedus:
- closes https://github.com/Shopify/spoom/issues/990

To ensure `T::Generic` and `T::Helpers` don't [conflict with user defined `T` constants](https://github.com/Shopify/nuevo-protobuf/pull/1095), we can fully qualify inserted `T::Generic`/`Helpers` to `::T::`

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
@@ -234,9 +234,8 @@ module Spoom
             # Only translate (and `extend ::T::Helpers`) when there's at least one *known* class
             # annotation. A node with only unknown annotations (e.g. `@private`) is left untouched.
             if comments.class_annotations.any?
-              unless already_extends?(node, /^(::)?T::Helpers$/)
-                extend_with("::T::Helpers", into: node, at: insert_pos)
-              end
+              find_and_remove_existing_extend(node, /^(::)?T::Helpers$/)
+              extend_with("::T::Helpers", into: node, at: insert_pos)
 
               comments.annotations.reverse_each do |annotation|
                 content = case annotation.string
@@ -286,9 +285,8 @@ module Spoom
                   next
                 end
 
-                unless already_extends?(node, /^(::)?T::Generic$/)
-                  extend_with("::T::Generic", into: node, at: insert_pos)
-                end
+                find_and_remove_existing_extend(node, /^(::)?T::Generic$/)
+                extend_with("::T::Generic", into: node, at: insert_pos)
 
                 type_params.each do |type_param|
                   type_member = "#{type_param.name} = type_member"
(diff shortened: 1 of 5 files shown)


lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -231,12 +231,11 @@ def apply_class_annotations(node)
...
             end
 
-            # Only translate (and `extend T::Helpers`) when there's at least one *known* class
+            # Only translate (and `extend ::T::Helpers`) when there's at least one *known* class
             # annotation. A node with only unknown annotations (e.g. `@private`) is left untouched.
             if comments.class_annotations.any?
-              unless already_extends?(node, /^(::)?T::Helpers$/)
-                extend_with("T::Helpers", into: node, at: insert_pos)
-              end
+              find_and_remove_existing_extend(node, /^(::)?T::Helpers$/)
```
amomchilov: I'm not sure if we should try to fix a _pre-existing_ `extend T::Helpers` like this.

It's a nice feature in theory, but narrow. There are many other explicit uses of `T` that would be problematic (`T.untyped`, `T::Boolean`, `T.any`, ...), which we couldn't patch up as easily/reliably. I'm inclined err of inaction, and require people fix their own explicit usages of `T`.

I suppose the distinction is: _who wrote the `T`?_

If our rewriter wrote it, we should be good stewards and `::T` it. If devs wrote it explicitly in their own source, they should choose `::T` if they want it (or rename their generic arg to not be called `T`).

amomchilov: @Morriar wdyt?

dejmedus: Makes sense! I can drop those commits and open up something separately if we decide we want the behaviour later on. In that case I think it would make sense to remove `already_extends?` (rather than fixing) and just allow, potentially, a duplicate `extends` 

<details>
<summary>RBS + pre-existing extends</summary>

```sh
# bin/console 
source = <<~RUBY
  # typed: true

  # @final
  #: [T]
  class Box
    extend T::Helpers
  end
RUBY

puts Spoom::Sorbet::Translate.rbs_comments_to_sorbet_sigs(
  source,
  file: "example.rb",
)
```

```rb
# typed: true

class Box
  extend ::T::Helpers # spoom inserted (still needed, otherwise we `extend` after `final!` unless we insert below)

  final!

  extend ::T::Generic

  T = type_member

  extend T::Helpers # source 
end
```
</details>

amomchilov: In a similar vein, I think that's too narrow of a problem-solve to be worth it. It would make their `extend T::Helpers`, but other code like `T.nilable` still fail.

I think it's a simpler (more predictable) mental model for the developer if `#: [T]` always caused unqualified accesses to fail, consistently

dejmedus: Agreed:) I removed `already_extends?` so it will continue to behave as it has been (inserting without checking for pre-existing) and a `[T]` + existing `T::Helpers/Generic` will error as expected

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -231,12 +231,10 @@ def apply_class_annotations(node)
...
             end
 
-            # Only translate (and `extend T::Helpers`) when there's at least one *known* class
+            # Only translate (and `extend ::T::Helpers`) when there's at least one *known* class
             # annotation. A node with only unknown annotations (e.g. `@private`) is left untouched.
             if comments.class_annotations.any?
-              unless already_extends?(node, /^(::)?T::Helpers$/)
-                extend_with("T::Helpers", into: node, at: insert_pos)
-              end
+              extend_with("::T::Helpers", into: node, at: insert_pos)
```
Morriar: Let's add a test doing this:

```rb
# @final
class Box
  extend T::Helpers
end
```

Do we add the extend anyway?

```rb
class Box
  extend T::Helpers
  extend ::T::Helpers
  final!
end
```

dejmedus: Yes:) Since `already_extends?` wasn't finding the node we've been inserting a duplicate even if there was a pre-existing. If we wanted `already_extends?` fixed instead of removed we'd need some logic to insert in the correct place (after the extends). Your example ends up as:

```
# typed: true

class Box
  extend ::T::Helpers # spoom inserted 

  final!

  extend T::Helpers # pre-existing
end
```

Morriar: Let's add this as a test to show the behavior is expected

dejmedus: Added [here](https://github.com/Shopify/spoom/pull/1001/changes#diff-f9a0f81ea7a119caba847b7c157e0e2eb0e7d9a0538d2a8a8dab65b1f00502b5R654):)

Morriar (approved):

lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
dejmedus: I moved the `already_extends?` removal out of this PR to #1014 so it can be attached to its own [issue](https://github.com/Shopify/spoom/issues/1013), but I've left the test here as behaviour is the same

PR #1017, opened 2026-08-18
Raise when type parameters collide during sig to RBS translation

### LINKED ISSUE #1012: srb sigs translate: type parameter shadows a type name used in the same signature
### Summary

`spoom srb sigs translate` converts Sorbet `sig` blocks into inline RBS comments (`#:`). When a `sig` declares a method type parameter whose name matches a type name referenced in the same signature, the emitted RBS is wrong: the type variable shadows whatever else that name referred to.

In Sorbet's `sig` DSL, `T.type_parameter(:Foo)` lives in a namespace of its own, separate from constants and from class-level type members. In RBS it does not — a type variable identifier is resolved ahead of anything else that could occupy the same type position. So every occurrence of that name in the translated signature binds to the type variable.

Three collision classes are affected. **Only the first fails loudly. The other two silently delete a type error.**

> Every reproduction below has a **playground link** carrying `?arg=--enable-experimental-rbs-comments&arg=--parser=prism`. Both flags are required — without them Sorbet ignores every `#:` comment and reports a misleading `7017` ("does not have a `sig`") instead of the real result. The same two flags are needed in `sorbet/config` to reproduce locally.

---

### Class 1 — collides with a user constant

```ruby
# typed: strict
class Store
  extend T::Sig

  module Credential
    module Multiple; end
  end

  sig do
    type_parameters(:Credential)
      .params(type: T.all(T::Module[T.type_parameter(:Credential)], T::Module[Credential::Multiple]))
      .returns(T::Array[T.type_parameter(:Credential)])
  end
  def find(type)
    []
  end
end
```

▶ [Run in Sorbet playground](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Store%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20module%20Credential%0A%20%20%20%20module%20Multiple%3B%20end%0A%20%20end%0A%0A%20%20sig%20do%0A%20%20%20%20type_parameters(%3ACredential)%0A%20%20%20%20%20%20.params(type%3A%20T.all(T%3A%3AModule%5BT.type_parameter(%3ACredential)%5D%2C%20T%3A%3AModule%5BCredential%3A%3AMultiple%5D))%0A%20%20%20%20%20%20.returns(T%3A%3AArray%5BT.type_parameter(%3ACredential)%5D)%0A%20%20end%0A%20%20def%20find(type)%0A%20%20%20%20%5B%5D%0A%20%20end%0Aend%0A) — clean before translation

`spoom srb sigs translate --from rbi --to rbs` emits:

```ruby
  #: [Credential] (
  #|   (Module[Credential] & Module[Credential::Multiple]) type
  #| ) -> Array[Credential]
```

▶ [Run in Sorbet playground](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Store%0A%20%20module%20Credential%0A%20%20%20%20module%20Multiple%3B%20end%0A%20%20end%0A%0A%20%20%23%3A%20%5BCredential%5D%20(%0A%20%20%23%7C%20%20%20(Module%5BCredential%5D%20%26%20Module%5BCredential%3A%3AMultiple%5D)%20type%0A%20%20%23%7C%20)%20-%3E%20Array%5BCredential%5D%0A%20%20def%20find(type)%0A%20%20%20%20%5B%5D%0A%20%20end%0Aend%0A) — `3550` + `7017`

`srb tc`, with the caret on the `::`:

```
Failed to parse RBS signature (comma delimited type list is expected)  https://srb.help/3550
The method `find` does not have a `sig`                                https://srb.help/7017
```

**It is the name, not the parenthesised intersection.** Changing one variable at a time:

- Remove the parentheses, keep the name → still fails, same `3550`. ▶ [playground](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Store%0A%20%20module%20Credential%0A%20%20%20%20module%20Multiple%3B%20end%0A%20%20end%0A%0A%20%20%23%3A%20%5BCredential%5D%20(%0A%20%20%23%7C%20%20%20Module%5BCredential%5D%20%26%20Module%5BCredential%3A%3AMultiple%5D%20type%0A%20%20%23%7C%20)%20-%3E%20Array%5BCredential%5D%0A%20%20def%20find(type)%0A%20%20%20%20%5B%5D%0A%20%20end%0Aend%0A)
- Keep the parentheses, rename the type parameter to `Cred` → passes, zero errors. ▶ [playground](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Store%0A%20%20module%20Credential%0A%20%20%20%20module%20Multiple%3B%20end%0A%20%20end%0A%0A%20%20%23%3A%20%5BCred%5D%20(%0A%20%20%23%7C%20%20%20(Module%5BCred%5D%20%26%20Module%5BCredential%3A%3AMultiple%5D)%20type%0A%20%20%23%7C%20)%20-%3E%20Array%5BCred%5D%0A%20%20def%20find(type)%0A%20%20%20%20%5B%5D%0A%20%20end%0Aend%0A)

This class fails loudly only because the shadowed constant is *dereferenced*. When it is merely mentioned, the same collision goes silent — see Class 1b.

### Class 1b — same, but the constant is only mentioned

```ruby
sig do
  type_parameters(:Credential)
    .params(x: T.type_parameter(:Credential), y: Credential)
    .returns(T.type_parameter(:Credential))
end
def find(x, y) = x
```

Translates to `#: [Credential] (Credential x, Credential y) -> Credential`, which parses fine.

```
BEFORE:  Store.new.find(1, 2)  →  Expected `Store::Credential` but found `Integer(2)`   ✓ caught
AFTER:   Store.new.find(1, 2)  →  No errors! Great job.                                  ✗ silent
         T.reveal_type(y)      →  T.type_parameter(:Credential)  (expected Store::Credential)
```

▶ [playground: before](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Store%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20module%20Credential%3B%20end%0A%0A%20%20sig%20do%0A%20%20%20%20type_parameters(%3ACredential)%0A%20%20%20%20%20%20.params(x%3A%20T.type_parameter(%3ACredential)%2C%20y%3A%20Credential)%0A%20%20%20%20%20%20.returns(T.type_parameter(%3ACredential))%0A%20%20end%0A%20%20def%20find(x%2C%20y)%0A%20%20%20%20x%0A%20%20end%0Aend%0A%0AStore.new.find(1%2C%202)%0A) · [playground: after](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Store%0A%20%20module%20Credential%3B%20end%0A%0A%20%20%23%3A%20%5BCredential%5D%20(Credential%20x%2C%20Credential%20y)%20-%3E%20Credential%0A%20%20def%20find(x%2C%20y)%0A%20%20%20%20T.reveal_type(y)%0A%20%20%20%20x%0A%20%20end%0Aend%0A%0AStore.new.find(1%2C%202)%0A)

### Class 2 — collides with a class-level `type_member`

The shadowed entity here is a type member, not a plain constant:

```ruby
# typed: strict
class Box
  extend T::Sig
  extend T::Generic

  Elem = type_member

  sig do
    type_parameters(:Elem)
      .params(x: T.type_parameter(:Elem), y: Elem)
      .returns(T.type_parameter(:Elem))
  end
  def pick(x, y)
    x
  end
end
```

Translates to `#: [Elem] (Elem x, Elem y) -> Elem`. The method type parameter shadows the class type parameter:

```
BEFORE:  Box[String].new.pick(1, 2)  →  Expected `String` but found `Integer(2)`   ✓ caught
AFTER:   Box[String].new.pick(1, 2)  →  No errors! Great job.                       ✗ silent
         T.reveal_type(y)            →  T.type_parameter(:Elem) (of Box#pick)
```

▶ [playground: before](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Box%0A%20%20extend%20T%3A%3ASig%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20Elem%20%3D%20type_member%0A%0A%20%20sig%20do%0A%20%20%20%20type_parameters(%3AElem)%0A%20%20%20%20%20%20.params(x%3A%20T.type_parameter(%3AElem)%2C%20y%3A%20Elem)%0A%20%20%20%20%20%20.returns(T.type_parameter(%3AElem))%0A%20%20end%0A%20%20def%20pick(x%2C%20y)%0A%20%20%20%20x%0A%20%20end%0Aend%0A%0ABox%5BString%5D.new.pick(1%2C%202)%0A) · [playground: after](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Box%0A%20%20extend%20T%3A%3AGeneric%0A%0A%20%20Elem%20%3D%20type_member%0A%0A%20%20%23%3A%20%5BElem%5D%20(Elem%20x%2C%20Elem%20y)%20-%3E%20Elem%0A%20%20def%20pick(x%2C%20y)%0A%20%20%20%20T.reveal_type(y)%0A%20%20%20%20x%0A%20%20end%0Aend%0A%0ABox%5BString%5D.new.pick(1%2C%202)%0A)

### Class 3 — collides with a builtin

```ruby
sig do
  type_parameters(:String)
    .params(a: T.type_parameter(:String), b: String)
    .returns(T.type_parameter(:String))
end
def go(a, b) = a
```

Translates to `#: [String] (String a, String b) -> String`. The type variable swallows the builtin:

```
BEFORE:  Q.new.go(1, 2)   →  Expected `String` but found `Integer(2)`   ✓ caught
AFTER:   Q.new.go(1, 2)   →  No errors! Great job.                       ✗ silent
         T.reveal_type(b) →  T.type_parameter(:String) (of Q#go)
```

▶ [playground: before](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Q%0A%20%20extend%20T%3A%3ASig%0A%0A%20%20sig%20do%0A%20%20%20%20type_parameters(%3AString)%0A%20%20%20%20%20%20.params(a%3A%20T.type_parameter(%3AString)%2C%20b%3A%20String)%0A%20%20%20%20%20%20.returns(T.type_parameter(%3AString))%0A%20%20end%0A%20%20def%20go(a%2C%20b)%0A%20%20%20%20a%0A%20%20end%0Aend%0A%0AQ.new.go(1%2C%202)%0A) · [playground: after](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Q%0A%20%20%23%3A%20%5BString%5D%20(String%20a%2C%20String%20b)%20-%3E%20String%0A%20%20def%20go(a%2C%20b)%0A%20%20%20%20T.reveal_type(b)%0A%20%20%20%20a%0A%20%20end%0Aend%0A%0AQ.new.go(1%2C%202)%0A)

---

### Expected

Detect the collision and refuse to translate that signature, reporting it — rather than emitting a signature that means something other than the input.

The check should fire when the type parameter name matches **any type name referenced in the same signature**, whether that resolves to a user constant, a builtin, or an in-scope `type_member` / `type_template`.

The *"referenced in the same signature"* qualifier is load-bearing in all three classes. A type parameter named `Credential` in a signature that never mentions `Credential` shadows nothing and is harmless, so the check should not fire on the name alone. That keeps this a real error rather than a nuisance.

**Please don't auto-rename the type parameter.** Renaming is safe in principle — a method type parameter name is local to its signature, and callers are observationally unaffected — but it is the wrong default here:

- It silently rewrites an identifier the author chose, which is visible in Sorbet's own diagnostics (`T.type_parameter(:Credential) (of Store#find)`) and to every future reader of the file.
- A renamer has to rewrite *every* type-variable occurrence and *no* constant occurrence in the same signature. Getting that boundary wrong emits a signature that parses cleanly and means something else — exactly the silent-mistyping failure this issue is about. That is a poor trade for a construct that occurred once in a 19,197-file translation.

Spoom already has precedent for both halves of this: `next if sig.is_abstract && !@translate_abstract_methods` skips what it can't translate faithfully, and `overloads_strategy` accepts `:raise` as one of `[:translate_all, :translate_last, :raise]`.

Two reasonable granularities:

- **Per-file** — skip the file, matching the existing `rescue RBI::Error` behaviour exactly. The collision check just needs to raise; the existing rescue reports it.
- **Per-signature** — leave that one `sig` block, translate the rest of the file, warn. Nicer output, slightly more work.

Either way it should be non-fatal. A hard abort would mean one bad signature kills a 110k-file migration, which is worse than the bug.

### Proposed error message

Following the style of `raise Error, "Method \`#{name}\` at #{location} has multiple overloaded signatures"`:

```
Can't translate the signature for `Store#find` at store.rb:9.

The type parameter `Credential` collides with the constant `Credential::Multiple`
referenced in the same signature. Sorbet's `sig` DSL keeps these in separate
namespaces; RBS does not, so the type variable would shadow the constant and
silently change what the signature means.

Left as a `sig` block. Rename the type parameter to something that doesn't
collide, then re-run.
```

The message should name **what** was shadowed, since it varies across the three classes — a user shown only the type parameter name would be baffled by the builtin case.

One wording note: the existing CLI rescue hardcodes `say_warning("Can't parse #{file}: ...")`, which would report a collision as a parse failure. This needs either a distinct rescue arm or a separate error class.

### Optional escape hatch

If maintainers want renaming available for anyone running a fully automatic large migration, the `overloads_strategy` precedent suggests a strategy option with `:raise` as the default. The objection above is to renaming being the default, not to it existing.

### Impact

Found while translating Shopify's Ruby monorepo (19,197 files rewritten). Class 1 accounted for 2 of that run's 14 remaining errors.

The error count understates it. Classes 1b, 2 and 3 produce no diagnostic at all — a signature that previously caught a type error stops catching it, and the diff shows a signature being faithfully translated. Those would not appear in any error count, including the one above, so the true incidence is unknown.

### Environment

spoom 1.8.7, sorbet / sorbet-static 0.6.13386, rbs 4.1.1, prism 1.9.0, Ruby 4.0.

### Related

#990 — "Generic types called `T` can lead to `TypeError`". Same shadowing family, but the opposite direction (RBS→RBI, runtime `NameError` from an inserted `extend T::Generic`), and its fix does **not** transfer here. #990 is fixed by fully qualifying the constant; that only half works for this issue:

| Emitted reference | Result | |
|---|---|---|
| `Credential::Multiple` — collision in first segment | parse error `3550` | ▶ [playground](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Store%0A%20%20module%20Credential%0A%20%20%20%20module%20Multiple%3B%20end%0A%20%20end%0A%0A%20%20%23%3A%20%5BCredential%5D%20(%0A%20%20%23%7C%20%20%20(Module%5BCredential%5D%20%26%20Module%5BCredential%3A%3AMultiple%5D)%20type%0A%20%20%23%7C%20)%20-%3E%20Array%5BCredential%5D%0A%20%20def%20find(type)%0A%20%20%20%20%5B%5D%0A%20%20end%0Aend%0A) |
| `::Shop::Credential::Multiple` — collision mid-path | resolves correctly | ▶ [playground](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Shop%0A%20%20module%20Credential%0A%20%20%20%20module%20Multiple%3B%20end%0A%20%20end%0A%0A%20%20%23%3A%20%5BCredential%5D%20((Module%5BCredential%5D%20%26%20Module%5B%3A%3AShop%3A%3ACredential%3A%3AMultiple%5D)%20type)%20-%3E%20Array%5BCredential%5D%0A%20%20def%20find(type)%0A%20%20%20%20T.reveal_type(type)%0A%20%20%20%20%5B%5D%0A%20%20end%0Aend%0A) |
| `::Store::Credential` — collision in final segment | silently becomes the type variable | ▶ [playground](https://sorbet.run/?arg=--enable-experimental-rbs-comments&arg=--parser=prism#%23%20typed%3A%20strict%0Aclass%20Store%0A%20%20module%20Credential%3B%20end%0A%0A%20%20%23%3A%20%5BCredential%5D%20(Credential%20x%2C%20%3A%3AStore%3A%3ACredential%20y)%20-%3E%20Credential%0A%20%20def%20test(x%2C%20y)%0A%20%20%20%20T.reveal_type(y)%0A%20%20%20%20x%0A%20%20end%0Aend%0A%0AStore.new.test(1%2C%202)%0A) |

Qualification clears the loud parse failure but not the silent mistyping.

That last row also looks like a Sorbet-side bug independent of spoom — a root-anchored path resolving to a type variable — and may warrant a separate upstream report.

#1010 — a different translator bug with the same shape of resolution.

### Workaround

Rename the type parameter so it doesn't collide, and leave a comment saying why, or the next person to touch the signature will "fix" the name back.

```ruby
# NOTE: the type parameter is `Cred`, not `Credential`: an RBS type variable shadows the
# constant of the same name, which would make `Credential::Multiple` unparseable.
#: [Cred] ((Module[Cred] & Module[Credential::Multiple]) type) -> Array[Cred]
```

Hashim1999164:
## Summary
* RBS type variables shadow constants and class type members that share the same name, so a translated sig can fail to parse or drop a type error
* When a method type parameter collides with a constant path or class type member in the same signature, raise an error that asks the user to rename the type parameter

Closes #1012

## Test plan
* bundle exec rake test


lib/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments.rb
```diff
@@ -388,6 +392,50 @@ def delete_extend_t_generics
...
 
+        #: (RBI::Sig) -> void
+        def rename_colliding_type_params!(sig)
+          return unless sig.type_params?
+
+          occupied = simple_type_names(sig) + class_type_member_names
+          mapping = {} #: Hash[String, String]
+
+          sig.type_params.each do |type_param|
+            next unless occupied.any? { |name| name == type_param || name.start_with?("#{type_param}::") }
```
Morriar: Should this compare against path segments too? #1012 calls out `::Store::Credential` as a silent collision. `simple_type_names` will collect the full name, so `type_parameters(:Credential)` with a root-qualified `::Store::Credential` reference doesn't match either condition here. We would still emit `[Credential] (... ::Store::Credential ...)`, which is the collision we're trying to avoid.

Let's add coverage for root-qualified / qualified paths before relying on this.

lib/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments.rb
```diff
@@ -388,6 +392,50 @@ def delete_extend_t_generics
...
+
+            mapping[type_param] = unique_type_param_name(type_param, occupied + mapping.values)
```
Morriar: Let's also reserve the existing method type parameters when choosing the replacement name.

With `type_parameters(:U, :UT)`, if `U` collides with a referenced constant, this currently picks `UT` because `UT` only appears as a type parameter, not in `occupied`. The emitted RBS would have duplicate type parameters and merge two independent variables.

```ruby
sig do
  type_parameters(:U, :UT)
    .params(x: T.type_parameter(:U), y: T.type_parameter(:UT), z: U)
    .returns(T.type_parameter(:U))
end
```

lib/spoom/sorbet/translate/sorbet_sigs_to_rbs_comments.rb
```diff
@@ -417,6 +465,86 @@ def rbs_print(indent, preserve_multiline_signatures:, &block)
...
+
+          #: (RBI::Type) -> void
+          def visit_type(type)
+            case type
+            when RBI::Type::Simple
+              visit_simple(type)
+            when RBI::Type::TypeParameter
+              visit_type_parameter(type)
+            when RBI::Type::Generic
+              type.params.each { |param| visit_type(param) }
```
Morriar: The collector only records `RBI::Type::Simple` names, but some referenced RBS names come from the container node itself. For example, `T::Array[T.type_parameter(:Array)]` references the builtin `Array`, but we only visit the generic params here, so `type_parameters(:Array)` won't be renamed and we'll still emit `#: [Array] ... Array[Array]`.

Same shape for `T::Hash`, `T::Module`, `T.class_of`, etc. Let's collect the wrapper/generic type names too, otherwise we can still emit a shadowed RBS signature.

jesse-shopify:
My concern with this is renaming a user's type name can lead to confusion if that name is surfaced later. I'm thinking a better strategy will be to raise an error suggesting the user rename to avoid the collision. That keeps the user in control and the names consistent.

Hashim1999164:
@jesse-shopify hi, i understand your concern. Can you please take it up from here. I wint be available upcoming few days. Thanks

Hashim1999164:
@jesse-shopify I switched this to raise instead of renaming. The error asks the user to rename the type parameter so the original names stay consistent and they stay in control.

Hashim1999164:
@jesse-shopify yeah that makes sense. dropped the rename approach. it raises now and tells you to rename the type param yourself so names stay consistent.

jesse-shopify (approved):

Hashim1999164:
@jesse-shopify CI was red on a Sorbet nilable thing for proc_returns / proc_bind. pushed a tiny fix that assigns them to locals before visit_type so the type narrows. should be green now

PR #1019, opened 2026-08-28
Place generic extend last

vinistock:
When we re-write RBS generic namespaces, we place the `extend T::Generic` first, in the very beginning of the body. This means that they appear last in the singleton class' ancestor chain, which makes it very easy to bypass Tapioca's patches for tracking generic type parameters.

```ruby
module Patch
  def [](*types)
  end
end

#: [Elem]
class Something
  extend Patch
end
```

That's enough to bypass the patch because `extend T::Generic` appears before `extend Patch`, which means the singleton ancestors are `[<Something>, <Patch>, <T::Generic>...]`. This breaks Tapioca RBI generation for generics.

Trying to ensure that Tapioca can always generate regardless of `[]` being overridden is not super straight forward, so I propose we flip the rewriter instead. This PR places `extend T::Generic` immediately after the last `extend` in a scope's body.

The idea is simply to change the insert position, which already fixes the issue, but then it also means we need to be a touch smarter about when to add line breaks in the body.

### DIFF
--- a/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
+++ b/lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
@@ -89,6 +89,31 @@ module Spoom
 
           private
 
+          #: (PrismTypes::anyScopeNode) -> Integer
+          def find_class_annotation_insert_pos(node)
+            # First try to find the last `extend` statement in the scope. This guarantees that we always place `extend
+            # T::Generic` after all the other extends, which makes it appear first in the ancestor chain
+            body = node.body
+
+            if body.is_a?(Prism::StatementsNode)
+              last_extend = body.body.reverse_each.find do |node|
+                node.is_a?(Prism::CallNode) && node.message == "extend" && !node.receiver
+              end
+
+              return last_extend.location.end_offset if last_extend
+            end
+
+            # If we don't find any extends, then place it immediately after the scope opening.
+            case node
+            when Prism::ClassNode
+              (node.superclass || node.constant_path).location.end_offset
+            when Prism::ModuleNode
+              node.constant_path.location.end_offset
+            when Prism::SingletonClassNode
+              node.expression.location.end_offset
+            end
+          end
+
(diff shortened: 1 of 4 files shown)


lib/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs/base_translator.rb
```diff
@@ -89,6 +89,31 @@ def visit_call_node(node)
...
 
+          #: (PrismTypes::anyScopeNode) -> Integer
+          def find_class_annotation_insert_pos(node)
+            # First try to find the last `extend` statement in the scope. This guarantees that we always place `extend
+            # T::Generic` after all the other extends, which makes it appear first in the ancestor chain
+            body = node.body
+
+            if body.is_a?(Prism::StatementsNode)
+              last_extend = body.body.reverse_each.find do |node|
+                node.is_a?(Prism::CallNode) && node.message == "extend" && !node.receiver
```
KaanOzkan, line 100: Technically we could have a receiver with `self.extend` right? We could also have a conditional call `extend Foo if x` which seems to be a `Prism::IfNode`. Rare to see though, so your call if you want to support it.

test/spoom/sorbet/translate/rbs_comments_to_sorbet_sigs_test.rb
```diff
@@ -622,6 +622,64 @@ class << self; extend ::T::Generic; A = type_member {{ fixed: ::T.class_of(Numer
...
+              end
+            RUBY
+
+            to_line_matched_format_for_machines: <<~RUBY,
+              # RBS_WRITTEN_ANNOTATION: [Elem]
+              class A
+                extend OtherThing; extend ::T::Generic; Elem = type_member
+              end
+            RUBY
+          )
```
KaanOzkan, line 649: It might be good to add a test for an `extend` with an inline comment. I'm working on another translator feature and inline comments caught me off guard.

vinistock:
We're going in a different direction and allowing for Tapioca gem extensions instead.

